PR #3, opened 2024-05-07
Discovering folders for pods

AngeCyp:
This PR offer a way to get all folders related to K8S's pods.
- Looking in the `/sys/fs/cgroup/kubepods.slice/` folder for every directory ending with `.slice` Each of them are added to look in if there is pods availables

- Minor change in cgroup's functions

### DIFF
--- a/plugin-k8s/src/cgroup_v2.rs
+++ b/plugin-k8s/src/cgroup_v2.rs
@@ -40,48 +40,25 @@ pub fn is_accessible_dir(path: &Path) -> bool {
     path.is_dir()
 }
 
-/// Retrieves the name of a cgroup without its prefix.
-///
-/// ## Example
-/// ```ignore
-/// use std::path::PathBuf;
-/// 
-/// let path = PathBuf::from("myPath/kubepods-burstable.slice/kubepods-burstable-podABCD");
-/// assert_eq!(retrieve_name(&path, &"kubepods-burstable-".to_string()).unwrap(),"podABCD");
-/// ```
-fn retrieve_name(path: &Path, prefix: &str) -> anyhow::Result<String> {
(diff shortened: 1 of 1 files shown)


plugin-k8s/src/cgroup_v2.rs
```diff
@@ -40,48 +40,24 @@ pub fn is_accessible_dir(path: &Path) -> bool {
...
+    let entries = fs::read_dir(root_directory_path)?;
     for entry in entries {
         let mut path = entry?.path();
         if path.is_dir() {
-            let _dir_name = path.file_name().expect("Impossible to write dir name");
-            let truncated_prefix = prefix.strip_suffix(".slice/").unwrap_or(&prefix);
-            let mut new_prefix = truncated_prefix.to_owned();
-            new_prefix.push_str("-");
-            let name = retrieve_name(&path, &new_prefix.to_owned())?;
+            let dir_name = path.file_name().expect("Impossible get dir name as String").to_os_string().into_string().map_err(|e| anyhow::Error::msg(format!("Conversion from Path to String error: {:?}", e)))?;
```
TheElectronWill, line 51: A suggestion that can be applied in multiple places, to make the code more readable:
```suggestion
            let dir_name = path.file_name().unwrap().to_str().ok_or(anyhow!("filename is not valid UTF-8: {}", path.display())?;
```
A simpler way to express the conversion of a path to a UTF-8 string would be `.to_str().ok_or(error)`.

Regarding the `file_name()`, according to the doc it only returns `None` when the path ends with `..`, which cannot happen here because you are listing the files of a directory. Hence I think `unwrap()` can be appropriate.

TheElectronWill (approved):

PR #5, opened 2024-05-28
Watch new pods created during measures

AngeC15:
Added a way to watch creation and deletion of pods.
Using inotify crate to add a watch on directories used by Kubernetes


plugin-k8s/src/lib.rs
```diff
@@ -43,7 +54,11 @@ impl AlumetPlugin for K8sPlugin {
 
     fn init(config: ConfigTable) -> anyhow::Result<Box<Self>> {
         let config = deserialize_config(config).context("invalid config")?;
-        Ok(Box::new(K8sPlugin { config }))
+
+        if let Ok(mut inotify_guard) = INOTIFY_VAR.lock() {
```
TheElectronWill, line 58: Instead of using a global variable with an `Arc<Mutex<Option<Whatever`, you should just store the value in the plugin structure.

plugin-k8s/Cargo.toml
```diff
@@ -13,3 +13,5 @@ reqwest = { version = "0.12", features = ["json"] }
 tokio = "1.37.0"
 serde = { version = "1.0.198", features = ["derive"] }
 humantime-serde = "1.1.1"
+inotify = "0.10.2"
+lazy_static = "1.4"
```
TheElectronWill, line 17:
```suggestion

```
You'll be able to remove this dependency once the code is cleaner (see comments on the lazy_static)

plugin-k8s/src/k8s_probe.rs
```diff
@@ -52,57 +66,115 @@ impl alumet::pipeline::Source for K8SProbe {
...
+                Ok(metrics_gathered) => {
+                    let diff_tot = match counter_tot.update(metrics_gathered.time_used_tot) {
+                        CounterDiffUpdate::FirstTime => None,
+                        CounterDiffUpdate::Difference(diff) | CounterDiffUpdate::CorrectedDifference(diff) => Some(diff),
+                    };
+                    let diff_usr = match counter_usr.update(metrics_gathered.time_used_user_mode) {
+                        CounterDiffUpdate::FirstTime => None,
+                        CounterDiffUpdate::Difference(diff) => Some(diff),
+                        CounterDiffUpdate::CorrectedDifference(diff) => Some(diff),
+                    };
```
TheElectronWill, line 81: I notice this pattern a lot (in other plugins too). Do you think that it would be a good thing to add a method `difference` to the `alumet` crate, which would turn your code into the following?
```rs
let diff_usr = counter_usr.update(metrics_gathered.time_used_user_mode).difference();
```

plugin-k8s/src/k8s_probe.rs
```diff
@@ -52,57 +66,115 @@ impl alumet::pipeline::Source for K8SProbe {
...
-                let p_usr: MeasurementPoint = MeasurementPoint::new(
-                    timestamp,
-                    self.metrics.time_used_user_mode,
-                    Resource::LocalMachine,
-                    consumer.clone(),
-                    value_usr as u64,
-                )
-                .with_attr("pod", AttributeValue::String(metrics.name.clone()));
-                measurements.push(p_usr);
+        }
```
TheElectronWill, line 134: What if the failure is temporary?
How about testing the metrics when the plugin starts, like the nvidia plugin?

plugin-k8s/src/k8s_probe.rs
```diff
@@ -52,57 +66,115 @@ impl alumet::pipeline::Source for K8SProbe {
...
+
+        let mut buffer = [0; 1024];
+        let events = if let Ok(mut inotify_guard) = INOTIFY_VAR.lock() {
+            if let Some(inotify) = inotify_guard.as_mut() {
+                match inotify.read_events(&mut buffer) {
+                    Ok(events) => Some(events),
+                    Err(error) if error.kind() == std::io::ErrorKind::WouldBlock => None,
+                    Err(_) => panic!("Error while reading events"),
```
TheElectronWill, line 142: this needs proper error management instead of a panic

TheElectronWill (changes requested):
I'm wondering whether this is the best approach to the problem.
Also, you need to replace all `println!` by `log::<something>` calls.

plugin-k8s/src/cgroup_v2.rs
```diff
@@ -67,20 +67,19 @@ fn list_metric_file_in_dir(root_directory_path: &Path) -> anyhow::Result<Vec<Cgr
     return Ok(vec_file_metric);
 }
 
-/// This function list all k8s pods availables, using 3 sub-directory to look in:
+/// This function list all k8s pods availables, using sub-directories to look in:
 /// For each subdirectory, we look in if there is a directory/ies about pods and we add it
 /// to a vector. All subdirectory are visited with the help of <list_metric_file_in_dir> function.
-pub fn list_all_k8s_pods_file(root_directory_str: &str) -> anyhow::Result<Vec<CgroupV2MetricFile>> {
+pub fn list_all_k8s_pods_file(root_directory_path: &PathBuf) -> anyhow::Result<Vec<CgroupV2MetricFile>> {
```
TheElectronWill, line 73:
```suggestion
pub fn list_all_k8s_pods_file(root_directory_path: &Path) -> anyhow::Result<Vec<CgroupV2MetricFile>> {
```

plugin-k8s/src/lib.rs
```diff
@@ -43,30 +50,128 @@ impl AlumetPlugin for K8sPlugin {
...
+        }
+        // struct JobDetector {
+        //     config_path: PathBuf,
+        //     time_spend_tot: TypedMetricId<u64>,
+        //     time_spend_usr: TypedMetricId<u64>,
+        //     time_spend_sys: TypedMetricId<u64>,
+        //     control_handle: ControlHandle,
+        //     plugin_name: String,
+        //     poll_interval: Duration,
+        // }
```
TheElectronWill, line 104: remove that?

plugin-k8s/src/lib.rs
```diff
@@ -43,30 +50,128 @@ impl AlumetPlugin for K8sPlugin {
...
+
+        impl EventHandler for PodDetector {
+            fn handle_event(&mut self, event: Result<Event, notify::Error>) {
+                // Handle_Event: Ok(Event { kind: Create(Folder), paths: ["/sys/fs/cgroup/kubepods.slice/kubepods-besteffort.slice/TESTTTTT"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None })
+                // Handle_Event: Ok(Event { kind: Remove(Folder), paths: ["/sys/fs/cgroup/kubepods.slice/kubepods-besteffort.slice/TESTTTTT"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None })
```
TheElectronWill, line 109: You can add a comment above those two comments to explain what it is.
For instance "the event, if successfull, looks like the following"

plugin-k8s/src/lib.rs
```diff
@@ -43,30 +50,128 @@ impl AlumetPlugin for K8sPlugin {
...
+     
+        Ok(())
+    }
+
 }
 
 impl Default for Config {
     fn default() -> Self {
+        let mut root_path = PathBuf::new();
+        root_path.push("/sys/fs/cgroup/kubepods.slice/");
```
TheElectronWill, line 172:
```suggestion
        let root_path = PathBuf::from("/sys/fs/cgroup/kubepods.slice/");
```

plugin-k8s/src/parsing_cgroupv2.rs
```diff
@@ -110,4 +110,4 @@ mod tests {
         assert_eq!(result.time_used_user_mode, 0 as u64);
         assert_eq!(result.time_used_system_mode, 0 as u64);
     }
-}
+}
```
TheElectronWill, line 113: please keep the trailing newline

TheElectronWill (changes requested):

plugin-k8s/src/lib.rs
```diff
@@ -43,30 +50,128 @@ impl AlumetPlugin for K8sPlugin {
...
+                        }
+
+                    }
+                } else if let Ok(Event {
+                    kind: EventKind::Remove(notify::event::RemoveKind::Folder),
+                    paths,
+                    ..
+                }) = event
+                {
+                    println!("suppression, path: {paths:?}");
```
TheElectronWill, line 146: println! should be replaced by log::debug! or similar

plugin-k8s/src/lib.rs
```diff
@@ -172,7 +155,7 @@ impl Default for Config {
         root_path.push("/sys/fs/cgroup/kubepods.slice/");
         Self {
             path: root_path,
-            poll_interval: Duration::from_secs(1), // 1Hz
+            poll_interval: Duration::from_millis(1), // 1Hz
```
TheElectronWill, line 158: Could you read that from the config?

TheElectronWill (approved):
very good

PR #12, opened 2024-06-19
Cgroups v1 and OAR2 Plugin Added

sofiaduqueg:


plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+            metrics: None,
+            watcher: None,
+        }))
+    }
+
+    fn start(&mut self, alumet: &mut AlumetStart) -> Result<(), anyhow::Error> {
+        let cpu_metric = alumet.create_metric::<u64>(
+            "cpu_time",
+            PrefixedUnit::nano(Unit::Second),
+            "Total CPU time consumed by the cgroup (in nanoseconds).",
```
AngeCyp, line 78:
```suggestion
            "Total CPU time consumed by the cgroup.",
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+        )?;
+        let memory_metric = alumet.create_metric::<u64>(
+            "memory_usage",
+            Unit::Unity,
+            "Total memory usage by the cgroup (in bytes).",
+        )?;
```
AngeCyp, line 84:
```suggestion
        let memory_metric = alumet.create_metric::<u64>(
            "memory_usage",
            Unit::Byte,
            "Total memory usage by the cgroup.",
        )?;
```

AngeCyp (changes requested):

plugin-oar2/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "plugin-oar2"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+alumet = { version = "0.4.1", path = "../alumet" }
+anyhow = "1.0.86"
+notify = { version = "6.1.1", default-features = false, features = ["macos_fsevent"] }
```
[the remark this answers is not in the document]
TheElectronWill: It does nothing on non-macOS targets, and selects a particular "backend" on macOS. When performing local tests, Sofia found that this one was better than the regular backend, which didn't report folder events properly on the macOS version she was using.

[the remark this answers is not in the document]
TheElectronWill: And the reason we need to specify a backend manually is because we use `default-features = false` in order to disable the `crossbeam-channel` feature, which causes interferences with tokio. See https://docs.rs/notify/latest/notify/#crossbeam-channel--tokio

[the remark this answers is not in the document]
AngeCyp: thank you for your explanations, it’s clear and ok for me

TheElectronWill:
LGTM, could you @sofiaduqueg rebase the branch on main?

TheElectronWill (approved):

titouanj (changes requested):
Sorry, my review wasn't published

titouanj (changes requested):
The plugin should be added to the workspace members in the main cargo.toml

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+                if let Ok(Event {
+                    kind: EventKind::Create(_),
+                    paths,
+                    ..
+                }) = event
+                {
+                    log::debug!("Paths: {paths:?}");
+                    for path in paths {
+                        if let Some(job_name) = path.file_name() {
+                            let job_name = job_name.to_str().unwrap();
```
titouanj, line 165:
```suggestion
                            let job_name = match job_name.to_str() {
                                Some(job_name) => job_name,
                                None => {
                                    log::error!("Unable to transform to str");
                                    continue;
                                },
                            };
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+        Ok(())
+    }
+
+    fn post_pipeline_start(&mut self, pipeline: &mut alumet::pipeline::runtime::RunningPipeline) -> anyhow::Result<()> {
+        let control_handle = pipeline.control_handle();
+        let config_path = self.config.path.clone();
+        let plugin_name = self.name().to_owned();
+
+        let metrics = self.metrics.clone().unwrap();
```
titouanj, line 139:
```suggestion
        let metrics = self.metrics.clone().ok_or(anyhow!("The error message you want"))?
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+            .with_attr("oar_job_id", self.job_id),
+        );
+
+        measurements.push(
+            MeasurementPoint::new(
+                timestamp,
+                self.memory_metric,
+                Resource::LocalMachine,
+                ResourceConsumer::ControlGroup {
+                    path: (self.memory_file_path.to_str().unwrap().to_owned().into()),
```
titouanj, line 277: I think it's better to leave the path empty or log an error than panicking

```suggestion
                    path: (self.memory_file_path.to_str().unwrap_or_default().to_owned().into()),
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+        let cpu_usage_u64 = cpu_usage.trim().parse::<u64>()?;
+        let memory_usage_u64 = memory_usage.trim().parse::<u64>()?;
+
+        measurements.push(
+            MeasurementPoint::new(
+                timestamp,
+                self.cpu_metric,
+                Resource::LocalMachine,
+                ResourceConsumer::ControlGroup {
+                    path: (self.cpu_file_path.to_str().unwrap().to_owned().into()),
```
titouanj, line 264: The same goes here

```suggestion
                    path: (self.cpu_file_path.to_str().unwrap_or_default().to_owned().into()),
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+                    ..
+                }) = event
+                {
+                    log::debug!("Paths: {paths:?}");
+                    for path in paths {
+                        if let Some(job_name) = path.file_name() {
+                            let job_name = job_name.to_str().unwrap();
+                            if job_name.chars().any(|c| c.is_numeric()) {
+                                let job_separated = job_name.split_once("_");
+                                let job_id = job_separated.context("Invalid oar cgroup").unwrap().1.parse().unwrap();
```
titouanj, line 168: I don't know what's best to manage these two unwrap, but the plugin should not panic on that. It could log an error and skip the current job or set a default value I don't know

titouanj (changes requested):

plugin-oar2/src/lib.rs
TheElectronWill: But `memory_file_path` has been obtained by the plugin before, it should in no case be invalid.

plugin-oar2/src/lib.rs
TheElectronWill: Creating a separate function that returns a result would be the best way to do it. It would allow to use `?`.

plugin-oar2/src/lib.rs
TheElectronWill, line 139:
```suggestion
        let metrics = self.metrics.take().expect("metrics is set in start()");
```

self.metrics cannot be None here

plugin-oar2/src/lib.rs
TheElectronWill: Maybe a separate function with `?` would be cleaner? I'm thinking about:

```rs
for path in paths {
    if let Err(e) = handle_event_on_path(path) {
        log::error!("Unable to handle event on {}: {e}", path.display());
    }
}
```

titouanj: Yeah I agree

plugin-oar2/src/lib.rs
titouanj: Yes you're right, but what should we do ? Keep using unwrap because a panic would be a real panic if happening, or logging the error and skipping the wrong/weird value ?

TheElectronWill: I'm tempted to keep unwrap (or use expect to provide an explanation of why it should never happen). Better detect the mistake early than propagating a weird value, isn't it?

Panics are caught by tokio anyway (i.e. it won't crash the whole pipeline).

titouanj: Yes, an expect would be good enough

plugin-oar2/src/lib.rs
```diff
@@ -98,7 +98,10 @@ impl AlumetPlugin for Oar2Plugin {
             let entry = entry?;
 
             let job_name = entry.file_name();
-            let job_name = job_name.into_string().ok().with_context(|| format!("Invalid oar username and job id."))?;
+            let job_name = job_name
+                .into_string()
+                .ok()
+                .with_context(|| format!("Invalid oar username and job id."))?;
```
TheElectronWill, line 104: please add job_name to the error message

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    fn post_pipeline_start(&mut self, pipeline: &mut alumet::pipeline::runtime::RunningPipeline) -> anyhow::Result<()> {
+        let control_handle = pipeline.control_handle();
+        let config_path = self.config.path.clone();
+        let plugin_name = self.name().to_owned();
+
+        let metrics = self
+            .metrics
+            .take()
+            .expect("Metrics are not available, metrics is set in start()");
```
TheElectronWill, line 145:
```suggestion
            .expect("Metrics should be initialized by start()");
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        measurements.push(
+            MeasurementPoint::new(
+                timestamp,
+                self.cpu_metric,
+                Resource::LocalMachine,
+                ResourceConsumer::ControlGroup {
+                    path: (self
+                        .cpu_file_path
+                        .to_str()
+                        .expect("cpu file path not available in str, it should never arrive")
```
TheElectronWill, line 281:
```suggestion
                        .expect("cpu_file_path should be valid UTF-8")
```

plugin-oar2/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        measurements.push(
+            MeasurementPoint::new(
+                timestamp,
+                self.memory_metric,
+                Resource::LocalMachine,
+                ResourceConsumer::ControlGroup {
+                    path: (self
+                        .memory_file_path
+                        .to_str()
+                        .expect("memory file path not available in str, it should never arrive")
```
TheElectronWill, line 299:
```suggestion
                        .expect("memory_file_path should be valid UTF-8")
```

TheElectronWill (changes requested):

TheElectronWill (approved):

PR #26, opened 2024-07-15
feat/Add a config field for the relay server plugin address

titouanj:
- fix(alumet): Add the name of the prefix symbols to UnitPrefix::from_str:
   When receiving a `MeasurementPoint` from a `RelayClientPlugin` by the
`RelayServerPlugin`, the `PrefixedUnit` may have a non empty prefix,
which would be equal to its `unique_name`. But the `from_str` was only
able to convert the `display_name` into the corresponding `UnitPrefix`.
It's now fixed by taking both names into account in the `from_str` match-case.

- feat(plugin-relay): Add a config field for the server listening address:
   It is now possible to give an address to listen on in the config
of the relay server plugin. The address can be an IP or any domain
name.

### DIFF
--- a/plugin-relay/src/server.rs
+++ b/plugin-relay/src/server.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
+    net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6, ToSocketAddrs},
     time::{Duration, UNIX_EPOCH},
 };
 
@@ -16,7 +16,7 @@ use alumet::{
     resources::{InvalidConsumerError, InvalidResourceError, Resource, ResourceConsumer},
     units::{PrefixedUnit, Unit},
 };
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use serde::{Deserialize, Serialize};
 use tonic::{transport::Server, Response, Status};
 
@@ -33,11 +33,14 @@ pub struct RelayServerPlugin {
 
 #[derive(Deserialize, Serialize)]
 struct Config {
+    /// Address to listen on.
+    /// The default value is ip6-localhost = `::1`.
+    ///
+    /// To listen all your network interfaces please use `0.0.0.0` or `::`.
+    address: String,
+
     /// Port on which to serve.
     port: u16,
-
-    /// If true, bind the server to an IPv4 socket instead of IPv6.
-    ipv4_only: bool,
 
     /// IPv6 scope id, for link-local addressing.
     ipv6_scope_id: Option<u32>,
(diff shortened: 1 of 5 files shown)


plugin-relay/src/server.rs
```diff
@@ -68,20 +71,38 @@ impl AlumetPlugin for RelayServerPlugin {
...
-        let addr: SocketAddr = match self.config.ipv4_only {
-            true => SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, self.config.port)),
-            false => SocketAddr::V6(SocketAddrV6::new(
-                Ipv6Addr::LOCALHOST,
+        // Retrieve the ip from the config using
+        // [the trait ToSocketAddrs](https://doc.rust-lang.org/stable/std/net/trait.ToSocketAddrs.html)
+        // of (String, u16) to use a raw address without any port.
+        let current_ip = match (self.config.address.clone(), self.config.port)
+            .to_socket_addrs()
+            .context("Parsing of the `address` param of the server relay plugin")?
```
TheElectronWill, line 85: Please include the config value in the error message. You can omit "server relay plugin", it's included by the logger based on the source file.

plugin-relay/src/server.rs
```diff
@@ -68,20 +71,38 @@ impl AlumetPlugin for RelayServerPlugin {
...
+            Some(socket) => socket.ip(),
+            _ => {
+                return Err(anyhow!("Could not retrieve any ip from {}", self.config.address));
+            }
+        };
+
+        // Once the IP has been parsed correctly, we need to build from scratch
+        // the SocketAddr to allow us to add `ipv6_scope_id` if needed.
+        let addr = match current_ip {
+            IpAddr::V4(ipv4) => SocketAddr::V4(SocketAddrV4::new(ipv4, self.config.port)),
```
TheElectronWill, line 97: If config.port is != socket.port() you could print a warning to explain that the port is ignored because it's in a separate config field, what do you think?

plugin-relay/src/server.rs
titouanj: Here is the error with the context:

```
 Caused by:
        0: Parsing of the `address` param of the server relay plugin: 127.0.0.1:50051
        1: failed to lookup address information: Name or service not known
```

plugin-relay/src/server.rs
titouanj: Well, when using to_socket_addrs on (String, u16), the string cannot have a port specified. It will raise the following error:

```
 Caused by:
        0: Parsing of the `address` param of the server relay plugin: 127.0.0.1:50051
        1: failed to lookup address information: Name or service not known
```

[See](https://doc.rust-lang.org/stable/std/net/trait.ToSocketAddrs.html)

TheElectronWill (approved):

PR #27, opened 2024-07-16
Feature/Add some configurations possibility

### LINKED ISSUE #23: Make config path... configurable
The title of this issue is not an oxymoron. The goal is to add a CLI argument to the agent that overrides the path of the TOML config file.

### LINKED ISSUE #24: Make socket path configurable
plugin: socket-control

Allow the path of the socket to be customized via the configuration file and/or command-line argument. This is required for deployment on Gricad with NixOS compose.

titouanj:
- Closes #23 with `--config`
- Closes #24 with the config part (i.e. not the cli)
- fix(workspace): Add plugin-oar2 to the rust workspace
- chore(plugin-relay): Bump the version from 0.1.0 to 0.2.0

### DIFF
--- a/app-agent/src/main.rs
+++ b/app-agent/src/main.rs
@@ -32,7 +32,7 @@ fn main() {
 
     // Build the measurement agent.
     let mut agent = AgentBuilder::new(plugins)
-        .config_path("alumet-config.toml")
+        .config_path(&args.config)
         .default_app_config(AppConfig::default())
         .build();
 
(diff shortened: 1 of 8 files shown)


app-agent/src/main.rs
```diff
@@ -32,7 +32,7 @@ fn main() {
 
     // Build the measurement agent.
     let mut agent = AgentBuilder::new(plugins)
-        .config_path("alumet-config.toml")
+        .config_path(args.config.clone())
```
TheElectronWill, line 35: Doesn't `&args.config` work?

TheElectronWill (commented):
Thanks!

app-agent/src/main.rs
titouanj: Yes indeed

PR #29, opened 2024-07-17
Stricter CI checks

TheElectronWill:
~~Based on #19 (to review, only look at the last 3 commits).~~

Rebased on main


TheElectronWill:
Tarpaulin is in trouble because some integration tests run commands... It fails with both backends :S
That will be for another PR.

plugin-k8s/src/parsing_cgroupv2.rs
```diff
@@ -24,8 +24,8 @@ impl FromStr for CgroupV2Metric {
             time_used_user_mode: 0,
             time_used_system_mode: 0,
         };
-        for (_i, line) in s.lines().enumerate() {
```
titouanj, line 27: What was this thing ? 🤣 @AngeCyp

Cargo.toml
```diff
@@ -18,3 +18,14 @@ members = [
...
+# To apply the following lint settings to a workspace member,
+# set "lints.workspace" to true in the member's Cargo.toml.
+[workspace.lints.clippy]
+redundant_field_names = "deny"
+needless_return = "deny"
+needless_borrow = "deny"
+comparison_to_empty = "deny"
+single_char_pattern = "deny"
+unused_enumerate_index = "deny"
+useless_format = "deny"
```
titouanj, line 31: Will it put the job in warning state if the is another warning from clippy which is not in this list ?

titouanj (approved):

titouanj:
> Tarpaulin is in trouble because some integration tests run commands... It fails with both backends :S That will be for another PR.

Doesn't this work ? https://github.com/xd009642/tarpaulin?tab=readme-ov-file#circleci

TheElectronWill:
> > Tarpaulin is in trouble because some integration tests run commands... It fails with both backends :S That will be for another PR.
> 
> Doesn't this work ? https://github.com/xd009642/tarpaulin?tab=readme-ov-file#circleci

It fails in local without the CI

Cargo.toml
TheElectronWill: it prints warning in the logs but there's no link with GitHub yet

plugin-k8s/src/parsing_cgroupv2.rs
AngeCyp: I really don't see any problem 🧑‍🦯 :/

PR #33, opened 2024-08-08
Feature/system and processes monitoring

TheElectronWill:


plugin-procfs/src/memory.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+fn convert_meminfo_to_bytes(value: u64, unit: &str) -> Option<u64> {
+    // For meminfo, "kB" actually means "kiB". See the doc of procfs.
+    match unit {
+        "B" => Some(value),
+        "kB" | "KiB" | "kiB" | "KB" => Some(value * 1024),
+        "mB" | "MiB" | "miB" | "MB" => Some(value * 1024 * 1024),
+        "gB" | "GiB" | "giB" | "GB" => Some(value * 1024 * 1024 * 1024),
```
titouanj, line 88: That's so weird

titouanj (approved):
Will you create the issues about every TODO ?

plugin-procfs/src/memory.rs
TheElectronWill: legacy issue 😅

PR #34, opened 2024-08-08
Added a test for source add in post_pipeline_start

AngeCyp:
This test should fail if there is an error about async when adding a new source in the post_pipeline_start function

### DIFF
--- a/alumet/Cargo.toml
+++ b/alumet/Cargo.toml
@@ -24,6 +24,7 @@ smallvec = { version = "1.13.2", features = ["union"] }
 tokio-util = "0.7.10"
 indoc = "2.0.5"
 thiserror = "1.0.62"
+humantime-serde = "1.1.1"
 
 # Dependencies for Linux builds only.
 [target.'cfg(target_os = "linux")'.dependencies]
--- a/alumet/tests/late_source_creation.rs
+++ b/alumet/tests/late_source_creation.rs
@@ -1,7 +1,7 @@
 
 use alumet::{agent::{AgentBuilder, AgentConfig}, measurement::{MeasurementAccumulator, MeasurementPoint, Timestamp}, metrics::TypedMetricId, pipeline::{elements::error::PollError, trigger::TriggerSpec}, plugin::{rust::{deserialize_config, serialize_config, AlumetPlugin}, AlumetPluginStart, AlumetPostStart, ConfigTable}, resources::{Resource, ResourceConsumer}, static_plugins, units::{PrefixedUnit, Unit}};
 use anyhow::Context;
-use serde::{Deserialize, Serialize};
+use tokio::{runtime::Runtime, time::timeout};
 use std::{thread, time::{self, Duration}};
 
 pub struct Metrics {
@@ -9,7 +9,6 @@ pub struct Metrics {
 }
 
 pub struct MyTestPluginLateMetricCreation {
-    config: Config,
     metrics: Option<Metrics>,
 }
 
@@ -18,12 +17,6 @@ struct MyTestSourcePlugin {
     value: TypedMetricId<u64>,
 }
 
-#[derive(Deserialize, Serialize)]
-struct Config {
-    /// Initial interval between two cgroup measurements.
-    #[serde(with = "humantime_serde")]
-    poll_interval: Duration,
-}
 
 impl AlumetPlugin for MyTestPluginLateMetricCreation {
     fn name() -> &'static str {
@@ -35,14 +28,12 @@ impl AlumetPlugin for MyTestPluginLateMetricCreation {
     }
 
     fn default_config() -> anyhow::Result<Option<ConfigTable>> {
-        let config = serialize_config(Config::default())?;
+        let config = serialize_config(Duration::from_secs(1))?;
         Ok(Some(config))
     }
 
-    fn init(config: ConfigTable) -> anyhow::Result<Box<Self>> {
-        let config = deserialize_config(config).context("invalid config")?;
+    fn init(_config: ConfigTable) -> anyhow::Result<Box<Self>> {
         Ok(Box::new(MyTestPluginLateMetricCreation {
-            config,
             metrics: None,
         }))
     }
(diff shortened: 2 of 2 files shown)


alumet/Cargo.toml
```diff
@@ -24,6 +24,7 @@ smallvec = { version = "1.13.2", features = ["union"] }
 tokio-util = "0.7.10"
 indoc = "2.0.5"
 thiserror = "1.0.62"
+humantime-serde = "1.1.1"
```
TheElectronWill, line 27: 1. This is only used in tests, it should be a dev dependency, not a regular dependency.
2. Why do you need it? It's only useful when there is a human that writes the config, but this is an automatic test.

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,123 @@
...
+impl AlumetPlugin for MyTestPluginLateMetricCreation {
+    fn name() -> &'static str {
+        "late_source_creation"
+    }
+
+    fn version() -> &'static str {
+        env!("CARGO_PKG_VERSION")
+    }
+
+    fn default_config() -> anyhow::Result<Option<ConfigTable>> {
```
TheElectronWill, line 37: IMO the test doesn't even need a configuration. You can arbitrary values :)

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,123 @@
...
+        let probe = MyTestSourcePlugin{
+            value:(self.metrics.as_ref().expect("Can't read byte_metric")).time_tot,
+        };
+
+        // Add the probe to the sources
+        control_handle
+            .add_source(
+                "x",
+                Box::new(probe),
+                TriggerSpec::at_interval(self.config.poll_interval),
```
TheElectronWill, line 75:
```suggestion
                TriggerSpec::at_interval(Duration::from_secs(1)), // arbitrary
```

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,123 @@
...
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            poll_interval: Duration::from_secs(1), // 1Hz
+        }
+    }
+}
+
+impl alumet::pipeline::Source for MyTestSourcePlugin {
+    fn poll(&mut self, measurements: &mut MeasurementAccumulator, timestamp: Timestamp) -> Result<(), PollError> {
```
TheElectronWill, line 91: Do you actually need to push measurements? If I remember correctly, the issue is about `control_handle.add_source`.

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,123 @@
...
+        .config_value(toml::Table::new())
+        .build();
+
+    // Stop the pipeline
+    let global_config = agent.default_config().unwrap();
+    let agent_config = AgentConfig::try_from(global_config).unwrap();
+    let agent = agent.start(agent_config).unwrap();
+    thread::sleep(time::Duration::from_secs(3));
+    agent.pipeline.control_handle().shutdown();
+    agent.wait_for_shutdown().unwrap();
```
TheElectronWill, line 121: Could you add a timeout here, to make sure that alumet shuts down at some point? It would prevent a regression in another part of the pipeline. To do this, you can use `tokio::timeout` with a current_thread runtime and `block_on`.

TheElectronWill (changes requested):

alumet/tests/late_source_creation.rs
AngeCyp: it was to be as close as possible to a real plugin

alumet/tests/late_source_creation.rs
AngeCyp: it was to be as close as possible to a real plugin

AngeCyp: but we don't need

alumet/tests/late_source_creation.rs
```diff
@@ -118,6 +102,14 @@ fn late_source_creation_test() {
...
     agent.pipeline.control_handle().shutdown();
-    agent.wait_for_shutdown().unwrap();
+    let rt = Runtime::new().unwrap();
+    let timeout_duration = Duration::from_secs(5);
+    let _result = rt.block_on(async {
+        timeout(timeout_duration, async {
+            agent.wait_for_shutdown().unwrap();
+        })
+        .await
+    });
```
TheElectronWill, line 103:
```suggestion
    let _result = rt.block_on(
        timeout(timeout_duration, async {
            agent.wait_for_shutdown().unwrap();
        })
    );
```

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,103 @@
...
+#[test]
+fn late_source_creation_test() {
+    /* This function test for a source add in the post_pipeline_start function */
+    let plugins = static_plugins![MyTestPluginLateMetricCreation];
+    
+    let agent = AgentBuilder::new(plugins)
+        .config_value(toml::Table::new())
+        .build();
+
+    // Stop the pipeline
```
TheElectronWill, line 90: This comment is a bit weird, lines 91 to 102 do much more than "stop the pipeline"

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,103 @@
+
+use alumet::{agent::{AgentBuilder, AgentConfig}, measurement::{MeasurementAccumulator, MeasurementPoint, Timestamp}, metrics::TypedMetricId, pipeline::{elements::error::PollError, trigger::TriggerSpec}, plugin::{rust::{deserialize_config, serialize_config, AlumetPlugin}, AlumetPluginStart, AlumetPostStart, ConfigTable}, resources::{Resource, ResourceConsumer}, static_plugins, units::{PrefixedUnit, Unit}};
+use anyhow::Context;
+use tokio::{runtime::Runtime, time::timeout};
+use std::{thread, time::{self, Duration}};
+
+pub struct Metrics {
```
TheElectronWill, line 7: this is no longer needed

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub time_tot: TypedMetricId<u64>,
+}
+
+pub struct MyTestPluginLateMetricCreation {
+    metrics: Option<Metrics>,
+}
+
+#[derive(Debug)]
+struct MyTestSourcePlugin {
+    value: TypedMetricId<u64>,
```
TheElectronWill, line 17: idem

TheElectronWill (changes requested):

alumet/tests/late_source_creation.rs
```diff
@@ -0,0 +1,103 @@
...
+impl alumet::pipeline::Source for MyTestSourcePlugin {
+    fn poll(&mut self, measurements: &mut MeasurementAccumulator, timestamp: Timestamp) -> Result<(), PollError> {
+        Ok(())
+    }
+}
+
+
+#[test]
+fn late_source_creation_test() {
+    /* This function test for a source add in the post_pipeline_start function */
```
TheElectronWill, line 83:
```suggestion

/// Test that adding a source after the pipeline has started works
#[test]
fn late_source_creation_test() {
```

TheElectronWill:
superseded by #35

PR #35, opened 2024-08-09
Summer core improvements

TheElectronWill:
Various improvements to the core of Alumet.
Note: most changes come from the documentation, perhaps it's easier to review this PR commit by commit.

### DIFF
--- a/alumet/tests/agent.rs
+++ b/alumet/tests/agent.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use alumet::{
     agent::{AgentBuilder, AgentConfig},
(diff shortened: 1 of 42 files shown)


alumet/tests/agent.rs
```diff
@@ -163,7 +163,7 @@ fn test_plugin_lifecycle() {
     let agent_config = AgentConfig::try_from(global_config).unwrap();
     let agent = agent.start(agent_config).unwrap();
     agent.pipeline.control_handle().shutdown();
-    agent.wait_for_shutdown().unwrap();
+    agent.wait_for_shutdown(Duration::from_secs(2)).unwrap();
```
AngeCyp, line 166: Did you tried other values ?

TheElectronWill: Yes, but the test shouldn't take long. I've chosen 2 instead of 1 in case the test runs on a slow machine.

PR #44, opened 2024-10-04
Feature: Add the first implementation of the k8s energy attribution

titouanj:
- added the first version of the k8s energy attribution plugin which cannot works if the collection frequency is higher than 1 Hz
- added cspell and fixed some typo in the same time


plugin-energy-attribution/src/transform.rs
```diff
@@ -0,0 +1,154 @@
...
+                        WrappedMeasurementValue::F64(fx) => cur_tot_time_f64 / tot_time_sum * fx,
+                        WrappedMeasurementValue::U64(ux) => cur_tot_time_f64 / tot_time_sum * (ux as f64),
+                    },
+                );
+
+                // Then we copy every attributes of the pod to the new MeasurementPoint.
+                point
+                    .clone()
+                    .attributes()
+                    .for_each(|(key, value)| new_m = new_m.clone().with_attr(key.to_owned(), value.clone()));
```
TheElectronWill, line 95: I won't request big changes because we already know what to improve for the next version, but I still have one comment here. Why do you clone `new_m` for each attribute? Also, you can use `with_attr_vec` or `with_attr_map` to add multiple attributes at once.

TheElectronWill (changes requested):

plugin-energy-attribution/src/transform.rs
titouanj: That's a really good question, and I don't remember the answer. Because we made this with @AngeCyp few months ago 😅

TheElectronWill (approved):
LGTM

PR #45, opened 2024-10-04
fix(plugin-energy-attribution): Change the cgroup metric name in the pre_pipeline_start

titouanj:
The name of the cgroup's metrics have been changed, so we need to change it too in the plugin-energy-atribution to make it works again...

`total_usage_usec` => `cgroup_cpu_usage_user`


TheElectronWill:
Ah, I missed it. Good catch!

We do need integration tests with each plugin to avoid that kind of problem.

TheElectronWill (approved):

PR #63, opened 2024-11-18
Feature/plugin_cgroupv2_k8s_memory

victoryeagle77:
- Factorization of code in poll function to simplify push of metric with the MeasurementPoint
- Correction with Cyprien of an issue to locate precisely the ".slice" folder content
- Adding data extraction from the "memory.stat" file in kubes to push following values :
- Anonyme active and inactive used memory
- Active and inactive file memory
- Interprocess communication shared memory
- Mapped files in memory
- Total memory = Anon memory + File memory + Shared memory + Mapped Files memory

### DIFF
--- a/.cspell/project-words.txt
+++ b/.cspell/project-words.txt
@@ -24,6 +24,7 @@ miri
 monitorable
 mpoint
 NVML
+pagetables
 PERFMON
 POWERCAP
 psys
--- a/plugin-cgroupv2/Cargo.toml
+++ b/plugin-cgroupv2/Cargo.toml
@@ -20,3 +20,7 @@ base64 = "0.22.1"
 
 [lints]
 workspace = true
+
+[dev-dependencies]
+tempfile = "3.5"
+mockito = "0.31"
--- a/plugin-cgroupv2/src/cgroupv2.rs
+++ b/plugin-cgroupv2/src/cgroupv2.rs
@@ -1,51 +1,72 @@
-use std::str::FromStr;
-
 use alumet::{
     metrics::{MetricCreationError, TypedMetricId},
     plugin::AlumetPluginStart,
     units::{PrefixedUnit, Unit},
 };
+use anyhow::{Context, Result};
+use std::str::FromStr;
 
 pub(crate) const CGROUP_MAX_TIME_COUNTER: u64 = u64::MAX;
 
 #[derive(Debug, PartialEq, Clone)]
-pub struct CgroupV2Metric {
-    pub name: String,
-    pub uid: String,
+pub struct CgroupMeasurements {
+    /// Name of a Kubernetes pod.
+    pub pod_name: String,
+    /// Unique identification of a Kubernetes pod.
+    pub pod_uid: String,
+    /// Resources isolation of a Kubernetes pod.
     pub namespace: String,
+    /// Kubernetes pod node.
     pub node: String,
-    pub time_used_tot: u64,
-    pub time_used_user_mode: u64,
-    pub time_used_system_mode: u64,
+    /// Total CPU usage time by the cgroup.
+    pub cpu_time_total: u64,
+    /// CPU in user mode usage time by the cgroup.
+    pub cpu_time_user_mode: u64,
+    /// CPU in system mode usage time by the cgroup.
+    pub cpu_time_system_mode: u64,
+    /// Anonymous used memory, corresponding to running process and various allocated memory.
+    pub memory_anonymous: u64,
+    // Files memory, corresponding to open files and descriptors.
+    pub memory_file: u64,
+    // Memory reserved for kernel operations.
+    pub memory_kernel: u64,
+    /// Memory used to manage correspondence between virtual and physical addresses.
+    pub memory_pagetables: u64,
 }
 
-impl FromStr for CgroupV2Metric {
+impl FromStr for CgroupMeasurements {
     type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let mut cgroup_struc_to_ret = CgroupV2Metric {
-            name: "".to_owned(),
-            uid: "".to_owned(),
+        let mut cgroup_struc_to_ret = CgroupMeasurements {
+            pod_name: "".to_owned(),
+            pod_uid: "".to_owned(),
             namespace: "".to_owned(),
             node: "".to_owned(),
-            time_used_tot: 0,
-            time_used_user_mode: 0,
-            time_used_system_mode: 0,
+            cpu_time_total: 0,
+            cpu_time_user_mode: 0,
+            cpu_time_system_mode: 0,
+            memory_anonymous: 0,
+            memory_file: 0,
+            memory_kernel: 0,
+            memory_pagetables: 0,
         };
+
         for line in s.lines() {
             let parts: Vec<&str> = line.split_ascii_whitespace().collect();
             if parts.len() >= 2 {
+                let value = parts[1]
+                    .parse::<u64>()
+                    .with_context(|| format!("Parsing of value : {}", parts[1]))?;
                 match parts[0] {
-                    "usage_usec" => {
-                        cgroup_struc_to_ret.time_used_tot = parts[1].parse::<u64>()?;
-                    }
-                    "user_usec" => {
-                        cgroup_struc_to_ret.time_used_user_mode = parts[1].parse::<u64>()?;
-                    }
-                    "system_usec" => {
-                        cgroup_struc_to_ret.time_used_system_mode = parts[1].parse::<u64>()?;
-                    }
-                    &_ => continue,
+                    "usage_usec" => cgroup_struc_to_ret.cpu_time_total = value,
+                    "user_usec" => cgroup_struc_to_ret.cpu_time_user_mode = value,
+                    "system_usec" => cgroup_struc_to_ret.cpu_time_system_mode = value,
+                    "anon" => cgroup_struc_to_ret.memory_anonymous = value,
+                    "file" => cgroup_struc_to_ret.memory_file = value,
+                    "kernel_stack" => cgroup_struc_to_ret.memory_kernel = value,
+                    "pagetables" => cgroup_struc_to_ret.memory_pagetables = value,
+                    _ => continue,
                 }
             }
         }
@@ -53,31 +74,83 @@ impl FromStr for CgroupV2Metric {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Eq, PartialEq)]
 pub struct Metrics {
-    pub time_used_tot: TypedMetricId<u64>,
-    pub time_used_user_mode: TypedMetricId<u64>,
-    pub time_used_system_mode: TypedMetricId<u64>,
+    /// Total CPU usage time by the cgroup.
+    pub cpu_time_total: TypedMetricId<u64>,
+    /// CPU in user mode usage time by the cgroup.
+    pub cpu_time_user_mode: TypedMetricId<u64>,
+    /// CPU in system mode usage time by the cgroup.
+    pub cpu_time_system_mode: TypedMetricId<u64>,
+    /// Anonymous used memory, corresponding to running process and various allocated memory.
+    pub memory_anonymous: TypedMetricId<u64>,
+    /// Files memory, corresponding to open files and descriptors.
+    pub memory_file: TypedMetricId<u64>,
+    /// Memory reserved for kernel operations.
+    pub memory_kernel: TypedMetricId<u64>,
+    /// Memory used to manage correspondence between virtual and physical addresses.
+    pub memory_pagetables: TypedMetricId<u64>,
+    /// Total memory used by cgroup.
+    pub memory_total: TypedMetricId<u64>,
 }
 
 impl Metrics {
+    /// Provides a information base to create metric before sending CPU and memory data,
+    /// with `name`, `unit` and `description` parameters.
+    ///
+    /// # Arguments
+    ///
+    /// * `alumet` - A AlumetPluginStart structure passed to plugins for the start-up phase.
+    ///
+    /// # Error
+    ///
+    ///  Return `MetricCreationError` when an error occur during creation a new metric.
     pub fn new(alumet: &mut AlumetPluginStart) -> Result<Self, MetricCreationError> {
-        let usec: PrefixedUnit = PrefixedUnit::micro(Unit::Second);
+        let usec = PrefixedUnit::micro(Unit::Second);
+
         Ok(Self {
-            time_used_tot: alumet.create_metric::<u64>(
+            // CPU cgroup data
+            cpu_time_total: alumet.create_metric::<u64>(
                 "cgroup_cpu_usage_total",
                 usec.clone(),
                 "Total CPU usage time by the cgroup",
             )?,
-            time_used_user_mode: alumet.create_metric::<u64>(
+            cpu_time_user_mode: alumet.create_metric::<u64>(
                 "cgroup_cpu_usage_user",
                 usec.clone(),
                 "CPU in user mode usage time by the cgroup",
             )?,
-            time_used_system_mode: alumet.create_metric::<u64>(
+            cpu_time_system_mode: alumet.create_metric::<u64>(
                 "cgroup_cpu_usage_system",
                 usec.clone(),
                 "CPU in system mode usage time by the cgroup",
+            )?,
+
+            // Memory cgroup data
+            memory_anonymous: alumet.create_metric::<u64>(
+                "cgroup_memory_anonymous",
+                Unit::Byte.clone(),
+                "Anonymous used memory, corresponding to running process and various allocated memory",
+            )?,
+            memory_file: alumet.create_metric::<u64>(
+                "cgroup_memory_file",
+                Unit::Byte.clone(),
+                "Files memory, corresponding to open files and descriptors",
+            )?,
+            memory_kernel: alumet.create_metric::<u64>(
+                "cgroup_memory_kernel_stack",
+                Unit::Byte.clone(),
+                "Memory reserved for kernel operations",
+            )?,
+            memory_pagetables: alumet.create_metric::<u64>(
+                "cgroup_memory_pagetables",
+                Unit::Byte.clone(),
+                "Memory used to manage correspondence between virtual and physical addresses",
+            )?,
+            memory_total: alumet.create_metric::<u64>(
+                "cgroup_memory_total",
+                Unit::Byte.clone(),
+                "Total memory used by cgroup",
             )?,
         })
     }
@@ -87,71 +160,119 @@ impl Metrics {
 mod tests {
     use super::*;
 
+    // Test `from_str` function with in extracted result,
+    // a negative value to test representation
     #[test]
-    fn test_parser() {
-        let str1 = format!(
-            "usage_usec 1111\n
-        user_usec 222222222222222222\n
-        system_usec 33\n
-        nr_periods 0\n
-        nr_throttled 0\n
-        throttled_usec 0"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str1).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 1111 as u64);
-        assert_eq!(result.time_used_user_mode, 222222222222222222 as u64);
-        assert_eq!(result.time_used_system_mode, 33 as u64);
+    fn test_signed_values() {
+        let str_cpu = "
+            usage_usec -10000
+            user_usec -20000
+            system_usec -30000";
 
-        let str2 = format!(
-            "nr_throttled 0\n
-        usage_usec 1111\n
-        system_usec 33"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str2).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 1111 as u64);
-        assert_eq!(result.time_used_user_mode, 0 as u64);
-        assert_eq!(result.time_used_system_mode, 33 as u64);
+        let str_memory = "
+            anon -10000
+            file -20000
+            kernel_stack -30000
+            pagetables -40000
+            percpu 890784
+            sock 16384
+            shmem 2453504
+            file_mapped -50000
+            ....";
 
-        let str3 = format!(
-            "
-        system_usec 33"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str3).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 0 as u64);
-        assert_eq!(result.time_used_user_mode, 0 as u64);
-        assert_eq!(result.time_used_system_mode, 33 as u64);
+        CgroupMeasurements::from_str(str_cpu).expect_err("ERROR : Signed value");
+        CgroupMeasurements::from_str(str_memory).expect_err("ERROR : Signed value");
+    }
 
-        let str4 = format!(
-            "usage_usec 1111\n
-        system_usec 33\n
-        user_usec 222222222222222222\n
-        throttled_usec 0"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str4).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 1111 as u64);
-        assert_eq!(result.time_used_user_mode, 222222222222222222 as u64);
-        assert_eq!(result.time_used_system_mode, 33 as u64);
+    // Test `from_str` function with in extracted result,
+    // a float or decimal value
+    #[test]
+    fn test_double_values() {
+        let str_cpu = "
+            usage_usec 10000.05
+            user_usec 20000.25
+            system_usec 30000.33";
 
-        let str5 = format!("");
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str5).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 0 as u64);
-        assert_eq!(result.time_used_user_mode, 0 as u64);
-        assert_eq!(result.time_used_system_mode, 0 as u64);
+        let str_memory = "
+            anon 10000.05
+            file 20000.25
+            kernel_stack 30000.33
+            pagetables 124325768932.56";
 
-        let str6 = format!(
-            "nr_periods 0\n
-        nr_throttled 0\n
-        throttled_usec 0"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str6).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 0 as u64);
-        assert_eq!(result.time_used_user_mode, 0 as u64);
-        assert_eq!(result.time_used_system_mode, 0 as u64);
+        CgroupMeasurements::from_str(str_cpu).expect_err("ERROR : Decimal value");
+        CgroupMeasurements::from_str(str_memory).expect_err("ERROR : Decimal value");
+    }
+
+    // Test `from_str` function with in extracted result,
+    // a null, empty or incompatible string
+    #[test]
+    fn test_invalid_values() {
+        let str_cpu = "
+            usage_usec !#⚠
+            user_usec
+            system_usec -123abc";
+
+        let str_memory = "
+            anon !#⚠
+            file
+            pagetables -123abc
+            ...";
+
+        CgroupMeasurements::from_str(str_cpu).expect_err("ERROR : Incompatible value");
+        CgroupMeasurements::from_str(str_memory).expect_err("ERROR : Incompatible value");
+    }
+
+    // Test `from_str` function with in extracted result,
+    // an empty string
+    #[test]
+    fn test_empty_values() {
+        let str: &str = "";
+        let result = CgroupMeasurements::from_str(str).unwrap();
+        // Memory file str
+        assert_eq!(result.memory_anonymous, 0);
+        assert_eq!(result.memory_file, 0);
+        assert_eq!(result.memory_kernel, 0);
+        assert_eq!(result.memory_pagetables, 0);
+        // CPU file str
+        assert_eq!(result.cpu_time_total, 0);
+        assert_eq!(result.cpu_time_user_mode, 0);
+        assert_eq!(result.cpu_time_system_mode, 0);
+    }
+
+    // Test for calculating `mem_total` with structure parameters
+    #[test]
+    fn test_calc_mem() {
+        let result = CgroupMeasurements {
+            pod_name: "".to_owned(),
+            pod_uid: "test_pod_uid".to_owned(),
+            namespace: "test_pod_namespace".to_owned(),
+            node: "test_pod_node".to_owned(),
+            cpu_time_total: 64,
+            cpu_time_user_mode: 16,
+            cpu_time_system_mode: 32,
+            memory_anonymous: 1024,
+            memory_file: 256,
+            memory_kernel: 4096,
+            memory_pagetables: 512,
+        };
+
+        let expected = CgroupMeasurements {
+            pod_name: "".to_owned(),
+            pod_uid: "test_pod_uid".to_owned(),
+            namespace: "test_pod_namespace".to_owned(),
+            node: "test_pod_node".to_owned(),
+            cpu_time_total: 64,
+            cpu_time_user_mode: 16,
+            cpu_time_system_mode: 32,
+            memory_anonymous: 1024,
+            memory_file: 256,
+            memory_kernel: 4096,
+            memory_pagetables: 512,
+        };
+
+        assert_eq!(result, expected);
+
+        let mem_total = result.memory_anonymous + result.memory_file + result.memory_kernel + result.memory_pagetables;
+        assert_eq!(mem_total, 5888);
     }
 }
--- a/plugin-cgroupv2/src/k8s/plugin.rs
+++ b/plugin-cgroupv2/src/k8s/plugin.rs
@@ -208,7 +208,7 @@ impl AlumetPlugin for K8sPlugin {
                                 let file_memory = File::open(&path_memory)
                                     .with_context(|| format!("failed to open file {}", path_memory.display()))?;
 
-                                // CPU ressource consumer for cpu.stat file in cgroup
+                                // CPU resource consumer for cpu.stat file in cgroup
                                 let consumer_cpu = ResourceConsumer::ControlGroup {
                                     path: path_cpu
                                         .to_str()
@@ -216,7 +216,7 @@ impl AlumetPlugin for K8sPlugin {
                                         .to_string()
                                         .into(),
                                 };
-                                // Memory ressource consumer for memory.stat file in cgroup
+                                // Memory resource consumer for memory.stat file in cgroup
                                 let consumer_memory = ResourceConsumer::ControlGroup {
                                     path: path_memory
                                         .to_str()
--- a/plugin-cgroupv2/src/k8s/plugin.rs
+++ b/plugin-cgroupv2/src/k8s/plugin.rs
@@ -5,8 +5,9 @@ use alumet::{
         util::CounterDiff,
         AlumetPluginStart, AlumetPostStart, ConfigTable,
     },
+    resources::ResourceConsumer,
 };
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use gethostname::gethostname;
 use notify::{Event, EventHandler, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use serde::{Deserialize, Serialize};
@@ -14,6 +15,7 @@ use std::{fs::File, path::PathBuf, time::Duration};
 
 use crate::{
     cgroupv2::{Metrics, CGROUP_MAX_TIME_COUNTER},
+    is_accessible_dir,
     k8s::utils::get_pod_name,
 };
 
@@ -38,12 +40,11 @@ struct K8sConfig {
     poll_interval: Duration,
     kubernetes_api_url: String,
     hostname: String,
-
     /// Way to retrieve the k8s API token.
     token_retrieval: TokenRetrieval,
 }
 
-#[derive(Clone, Deserialize, Serialize)]
+#[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
 #[serde(rename_all = "lowercase")]
 pub enum TokenRetrieval {
     Kubectl,
@@ -73,10 +74,14 @@ impl AlumetPlugin for K8sPlugin {
         }))
     }
 
+    fn stop(&mut self) -> anyhow::Result<()> {
+        Ok(())
+    }
+
     fn start(&mut self, alumet: &mut AlumetPluginStart) -> anyhow::Result<()> {
-        let v2_used: bool = super::utils::is_accessible_dir(&PathBuf::from("/sys/fs/cgroup/"));
+        let v2_used = is_accessible_dir(&PathBuf::from("/sys/fs/cgroup/"))?;
         if !v2_used {
-            anyhow::bail!("Cgroups v2 are not being used!");
+            return Err(anyhow!("Cgroups v2 are not being used!"));
         }
         self.metrics = Some(Metrics::new(alumet)?);
 
@@ -95,11 +100,13 @@ impl AlumetPlugin for K8sPlugin {
             self.config.kubernetes_api_url.clone(),
             &Token::new(self.config.token_retrieval.clone()),
         )?;
-        //Add as a source each pod already present
+
+        // Add as a source each pod already present
         for metric_file in final_list_metric_file {
-            let counter_tmp_tot: CounterDiff = CounterDiff::with_max_value(crate::cgroupv2::CGROUP_MAX_TIME_COUNTER);
-            let counter_tmp_usr: CounterDiff = CounterDiff::with_max_value(crate::cgroupv2::CGROUP_MAX_TIME_COUNTER);
-            let counter_tmp_sys: CounterDiff = CounterDiff::with_max_value(crate::cgroupv2::CGROUP_MAX_TIME_COUNTER);
+            let counter_tmp_tot = CounterDiff::with_max_value(crate::cgroupv2::CGROUP_MAX_TIME_COUNTER);
+            let counter_tmp_usr = CounterDiff::with_max_value(crate::cgroupv2::CGROUP_MAX_TIME_COUNTER);
+            let counter_tmp_sys = CounterDiff::with_max_value(crate::cgroupv2::CGROUP_MAX_TIME_COUNTER);
+
             let probe = K8SProbe::new(
                 self.metrics.as_ref().expect("Metrics is not available").clone(),
                 metric_file,
@@ -113,14 +120,10 @@ impl AlumetPlugin for K8sPlugin {
         Ok(())
     }
 
-    fn stop(&mut self) -> anyhow::Result<()> {
-        Ok(())
-    }
-
     fn post_pipeline_start(&mut self, alumet: &mut AlumetPostStart) -> anyhow::Result<()> {
         let control_handle = alumet.pipeline_control();
 
-        let metrics: Metrics = self.metrics.clone().expect("Metrics is not available");
+        let metrics = self.metrics.clone().expect("Metrics is not available");
         let poll_interval = self.config.poll_interval;
         let kubernetes_api_url = self.config.kubernetes_api_url.clone();
         let hostname = self.config.hostname.to_owned();
@@ -153,24 +156,27 @@ impl AlumetPlugin for K8sPlugin {
                         for path in paths {
                             match path.extension() {
                                 None => {
-                                    // Case of no extension found --> I will not find cpu.stat file
+                                    // Case of no extension found --> I will not find cpu.stat or memory.stat file
                                     return Ok(());
                                 }
                                 Some(os_str) => match os_str.to_str() {
                                     Some("slice") => {
-                                        // Case of .slice found --> I will find cpu.stat file
+                                        // Case of .slice found --> I will find cpu.stat or memory.stat file
                                         log::debug!(".slice extension found, will continue");
                                     }
                                     _ => {
-                                        // Case of an other extension than .slice is found --> I will not find cpu.stat file
+                                        // Case of an other extension than .slice is found --> I will not find cpu.stat or memory.stat file
                                         return Ok(());
                                     }
                                 },
                             };
+
                             if let Some(pod_uid) = path.file_name() {
                                 let pod_uid = pod_uid.to_str().expect("Can't retrieve the pod uid value");
+
                                 // We open a File Descriptor to the newly created file
                                 let mut path_cpu = path.clone();
+                                let mut path_memory = path.clone();
                                 let full_name_to_seek = pod_uid.strip_suffix(".slice").unwrap_or(pod_uid);
                                 let parts: Vec<&str> = full_name_to_seek.split("pod").collect();
                                 let name_to_seek_raw = *(parts.last().unwrap_or(&full_name_to_seek));
@@ -195,22 +201,46 @@ impl AlumetPlugin for K8sPlugin {
                                     .with_context(|| "Block on failed returned an error")?;
 
                                 path_cpu.push("cpu.stat");
-                                let file = File::open(&path_cpu)
+                                let file_cpu = File::open(&path_cpu)
                                     .with_context(|| format!("failed to open file {}", path_cpu.display()))?;
+
+                                path_memory.push("memory.stat");
+                                let file_memory = File::open(&path_memory)
+                                    .with_context(|| format!("failed to open file {}", path_memory.display()))?;
+
+                                // CPU ressource consumer for cpu.stat file in cgroup
+                                let consumer_cpu = ResourceConsumer::ControlGroup {
+                                    path: path_cpu
+                                        .to_str()
+                                        .expect("Path to 'cpu.stat' must be valid UTF8")
+                                        .to_string()
+                                        .into(),
+                                };
+                                // Memory ressource consumer for memory.stat file in cgroup
+                                let consumer_memory = ResourceConsumer::ControlGroup {
+                                    path: path_memory
+                                        .to_str()
+                                        .expect("Path to 'memory.stat' must to be valid UTF8")
+                                        .to_string()
+                                        .into(),
+                                };
 
                                 let metric_file = CgroupV2MetricFile {
                                     name: name.to_owned(),
-                                    path: path_cpu,
-                                    file,
+                                    consumer_cpu,
+                                    file_cpu,
+                                    consumer_memory,
+                                    file_memory,
                                     uid: uid.to_owned(),
                                     namespace: namespace.to_owned(),
                                     node: node.to_owned(),
                                 };
 
-                                let counter_tmp_tot: CounterDiff = CounterDiff::with_max_value(CGROUP_MAX_TIME_COUNTER);
-                                let counter_tmp_usr: CounterDiff = CounterDiff::with_max_value(CGROUP_MAX_TIME_COUNTER);
-                                let counter_tmp_sys: CounterDiff = CounterDiff::with_max_value(CGROUP_MAX_TIME_COUNTER);
-                                let probe: K8SProbe = K8SProbe::new(
+                                let counter_tmp_tot = CounterDiff::with_max_value(CGROUP_MAX_TIME_COUNTER);
+                                let counter_tmp_usr = CounterDiff::with_max_value(CGROUP_MAX_TIME_COUNTER);
+                                let counter_tmp_sys = CounterDiff::with_max_value(CGROUP_MAX_TIME_COUNTER);
+
+                                let probe = K8SProbe::new(
                                     detector.metrics.clone(),
                                     metric_file,
                                     counter_tmp_tot,
@@ -261,6 +291,9 @@ impl AlumetPlugin for K8sPlugin {
 impl Default for K8sConfig {
     fn default() -> Self {
         let root_path = PathBuf::from("/sys/fs/cgroup/kubepods.slice/");
+        if !root_path.exists() {
+            log::warn!("Error : Path '{}' not exist.", root_path.display());
+        }
         Self {
             path: root_path,
             poll_interval: Duration::from_secs(1), // 1Hz
@@ -268,5 +301,62 @@ impl Default for K8sConfig {
             hostname: String::from(""),
             token_retrieval: TokenRetrieval::Kubectl,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use anyhow::Result;
+    use std::path::PathBuf;
+
+    // Create a fake plugin structure for k8s plugin
+    fn fake_k8s() -> K8sPlugin {
+        K8sPlugin {
+            config: K8sConfig {
+                path: PathBuf::from("/sys/fs/cgroup/kubepods.slice/"),
+                poll_interval: Duration::from_secs(1),
+                kubernetes_api_url: String::from("https://127.0.0.1:8080"),
+                hostname: String::from("test-hostname"),
+                token_retrieval: TokenRetrieval::Kubectl,
+            },
+            watcher: None,
+            metrics: None,
+        }
+    }
+
+    // Test `default_config` function of k8s plugin
+    #[test]
+    fn test_default_config() {
+        let result = K8sPlugin::default_config().unwrap();
+        assert!(result.is_some(), "result = None");
+
+        let config_table = result.unwrap();
+        let config: K8sConfig = deserialize_config(config_table).expect("Failed to deserialize config");
+
+        assert_eq!(config.path, PathBuf::from("/sys/fs/cgroup/kubepods.slice/"));
+        assert_eq!(config.poll_interval, Duration::from_secs(1));
+        assert_eq!(config.kubernetes_api_url, "https://127.0.0.1:8080");
+        assert_eq!(config.hostname, "");
+        assert_eq!(config.token_retrieval, TokenRetrieval::Kubectl);
+    }
+
+    // Test `init` function to initialize k8s plugin configuration
+    #[test]
+    fn test_init() -> Result<()> {
+        let config_table = serialize_config(K8sConfig::default())?;
+        let plugin = K8sPlugin::init(config_table)?;
+        assert_eq!(plugin.config.kubernetes_api_url, "https://127.0.0.1:8080");
+        assert!(plugin.metrics.is_none());
+        assert!(plugin.watcher.is_none());
+        Ok(())
+    }
+
+    // Test `stop` function to stop k8s plugin
+    #[test]
+    fn test_stop() {
+        let mut plugin = fake_k8s();
+        let result = plugin.stop();
+        assert!(result.is_ok(), "Stop should complete without errors.");
     }
 }
--- a/plugin-cgroupv2/src/k8s/probe.rs
+++ b/plugin-cgroupv2/src/k8s/probe.rs
@@ -1,24 +1,28 @@
 use alumet::{
     measurement::{AttributeValue, MeasurementAccumulator, MeasurementPoint, Timestamp},
     metrics::TypedMetricId,
-    pipeline::elements::error::PollError,
+    pipeline::elements::error::{PollError, PollRetry},
     plugin::util::CounterDiff,
     resources::{Resource, ResourceConsumer},
 };
-use anyhow::Result;
-
-use crate::cgroupv2::{CgroupV2Metric, Metrics};
+use anyhow::{Context, Result};
 
 use super::utils::{gather_value, CgroupV2MetricFile};
+use crate::cgroupv2::{CgroupMeasurements, Metrics};
 
 pub struct K8SProbe {
     pub cgroup_v2_metric_file: CgroupV2MetricFile,
     pub time_tot: CounterDiff,
     pub time_usr: CounterDiff,
     pub time_sys: CounterDiff,
-    pub time_used_tot: TypedMetricId<u64>,
-    pub time_used_user_mode: TypedMetricId<u64>,
-    pub time_used_system_mode: TypedMetricId<u64>,
+    pub cpu_time_tot: TypedMetricId<u64>,
+    pub cpu_time_user_mode: TypedMetricId<u64>,
+    pub cpu_time_system_mode: TypedMetricId<u64>,
+    pub memory_anon: TypedMetricId<u64>,
+    pub memory_file: TypedMetricId<u64>,
+    pub memory_kernel: TypedMetricId<u64>,
+    pub memory_pagetables: TypedMetricId<u64>,
+    pub memory_total: TypedMetricId<u64>,
 }
 
 impl K8SProbe {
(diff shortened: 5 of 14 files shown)


plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -79,79 +171,33 @@ impl Metrics {
...
-    #[test]
-    fn test_parser() {
-        let str1 = format!(
-            "usage_usec 1111\n
-        user_usec 222222222222222222\n
-        system_usec 33\n
-        nr_periods 0\n
-        nr_throttled 0\n
-        throttled_usec 0"
-        );
```
AngeCyp, line 99: Where are the previous tests ?

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -155,13 +165,23 @@ pub fn list_all_k8s_pods_file(
...
+        .with_context(|| format!("Unable to gather cgroup v2 metrics by reading file {}", file.name))?;
+    file.file_cpu.rewind()?;
+
+    file.file_memory
         .read_to_string(content_buffer)
         .with_context(|| format!("Unable to gather cgroup v2 metrics by reading file {}", file.name))?;
-    file.file.rewind()?;
-    let mut new_metric =
-        CgroupV2Metric::from_str(content_buffer).with_context(|| format!("failed to parse {}", file.name))?;
+    file.file_memory.rewind()?;
```
AngeCyp, line 178: Did you test that metrics are not merged and still readable by your parser ?

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -155,13 +165,23 @@ pub fn list_all_k8s_pods_file(
...
+
+    //println!("-------------- {:?}", content_buffer);
```
AngeCyp, line 180: to remove :)

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -373,132 +395,3 @@ pub async fn get_pod_name(
...
-    fn test_list_metric_file_in_dir() {
-        let tmp = std::env::temp_dir();
-        let root: std::path::PathBuf = tmp.join("test-alumet-plugin-k8s/kubepods-folder.slice/");
-        if root.exists() {
-            std::fs::remove_dir_all(&root).unwrap();
-        }
-        let burstable_dir = root.join("kubepods-burstable.slice/");
-        std::fs::create_dir_all(&burstable_dir).unwrap();
-
-        let a = burstable_dir.join("kubepods-burstable-pod32a1942cb9a81912549c152a49b5f9b1.slice/");
```
AngeCyp, line 406: Where are they ?

AngeCyp (changes requested):

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -17,11 +41,27 @@ pub struct CgroupV2Metric {
...
+    pub file_mapped_mem: u64,
+    pub total_mem: u64
 }
 
 impl FromStr for CgroupV2Metric {
     type Err = anyhow::Error;
-
+    /// # Function
+    /// 
+    /// * `from_str` - Function provides functionality to parse a string by whitespaces places.
```
titouanj, line 55: I don't think it is that useful to specify in a function doc that this is a function

titouanj (commented):
We need to talk about the documentation/comments convention @TheElectronWill

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -8,6 +13,25 @@ use alumet::{
...
+/// * `node` - Kubernetes pod node
+/// * `time_used_tot` - Total CPU usage time by the cgroup
+/// * `time_used_user_mode` - CPU in user mode usage time by the cgroup
+/// * `time_used_system_mode` - CPU in system mode usage time by the cgroup
+/// * `anon_used_mem` - Anonyme used memory, corresponding to running process and various allocated memory
+/// * `file_mem` - Files memory, corresponding to open files and descriptors
+/// * `shared_mem` - Interprocess communication shared memory
+/// * `file_mapped_mem` - Mapped files in memory
+/// * `total_mem` - Total memory used by cgroup
+///
```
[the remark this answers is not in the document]
TheElectronWill: Let's follow the Rust ecosystem's way of documenting. In particular:
- don't repeat the name of the struct at the beginning of the doc (it will be shown automatically)
- the first line should be a description, not a title (the title is automatically added by rustdoc)
- each field of the struct should be documented on its own, not in a list in the struct description

victoryeagle77:
Memory data for _plugin-cgroupv2_ for _k8s_ module added{color}
 * Factorization of code in _poll_ function to simplify push of metric with the _MeasurementPoint_
 * Correction with Cyprien of an issue to locate precisely the ".slice" folder content
 * Adding data extraction from the "memory.stat" file in kubes to push following values :
  -> Anonymous active and inactive used memory.
  -> Active and inactive file memory.
  -> Kernel memory allocated (npn) including (kernel_stack, pagetables, percpu, vmalloc, slab)
       in addition to other kernel memory use cases.
  -> Memory allocated for page tables
  -> With Guillaume, we determine the following formula to calculate cgroup memory :
       Total memory = Anon memory + File memory + kernel memory + Pagetables memory
 * Pull request and validation tests for commit on Alumet repository

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -12,55 +12,25 @@ use alumet::{
...
-    pub node: String,
-    pub time_used_tot: u64,
-    pub time_used_user_mode: u64,
-    pub time_used_system_mode: u64,
-    pub anon_used_mem: u64,
-    pub file_mem: u64,
-    pub shared_mem: u64,
-    pub file_mapped_mem: u64,
-    pub total_mem: u64,
+    pub name: String,               // Name of a Kubernetes pod
```
TheElectronWill, line 17:
```suggestion
    /// Name of a Kubernetes pod
    pub name: String,               
```
Here is how each field should be documented. In Rust, `//` is a regular comment (not in the doc), and `///` is a doc comment`. Every doc comment must be written just above the element that it documents.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -196,94 +147,82 @@ impl Metrics {
...
-            throttled_usec 0"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str6).unwrap();
+        let test_cgroup_files: String = format!("");
+        let result: CgroupV2Metric = CgroupV2Metric::from_str(&test_cgroup_files).unwrap();
         assert_eq!(result.name, "");
         assert_eq!(result.time_used_tot, 0 as u64);
         assert_eq!(result.time_used_user_mode, 0 as u64);
         assert_eq!(result.time_used_system_mode, 0 as u64);
+        assert_eq!(result.anon_used_mem, 0 as u64);
```
TheElectronWill, line 223: Is the `as u64` required?

plugin-cgroupv2/src/k8s/probe.rs
```diff
@@ -93,6 +93,19 @@ impl K8SProbe {
...
+    ///
+    /// # Parameters
+    ///
+    /// - `measurements` : An accumulator stores measured data points
+    /// - `timestamp` : Time recorded at the time of measurement
+    ///
+    /// # Return
+    ///
+    /// Success or and error which can occur during `Source::poll`
+    ///
```
TheElectronWill, line 108: Like structures, functions should be documented in a "direct" approach. We know it's a function, don't say it in the title.
Furthermore, `poll` is defined in `alumet::pipeline::Source`, there is no need to document the same thing in its implementation. If you have specific things to say about _this implementation of `poll`_, write them in the doc. Otherwise, I think that it's unnecessary to document it.

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -17,27 +17,21 @@ use std::{
...
-/// # Structure
-///
-/// Public structure `CgroupV2MetricFile` creating a log file containing cgroupv2 data about :
-/// - cpu usage
-/// - Memory usage
-///
 #[derive(Debug)]
 pub struct CgroupV2MetricFile {
-    /// Name of the pod.
+    /// Name of the pod
```
TheElectronWill, line 22: Actually, I liked the dot :)
It matches the Rust std style. [Example](https://doc.rust-lang.org/src/core/option.rs.html#571)

TheElectronWill (changes requested):

plugin-cgroupv2/Cargo.toml
```diff
@@ -20,3 +20,8 @@ base64 = "0.22.1"
 
 [lints]
 workspace = true
+
+[profile.release]
+opt-level = 1
+lto = true
+codegen-units = 1
```
TheElectronWill, line 27: Per se it's not a bad idea to enable more aggressive LTO, however:
- it should not be applied on a plugin (because it won't do anything)
- this modification should be in another PR
- the benefits should be demonstrated (how much do we gain in perf, how much do we loose in compilation time, what impact does it have on binary size?)

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -31,21 +49,77 @@ impl FromStr for CgroupV2Metric {
...
-                        cgroup_struc_to_ret.time_used_user_mode = parts[1].parse::<u64>()?;
+                // Parse value as f64
+                match parts[1].parse::<f64>() {
+                    Ok(value) => {
+                        // Check if value is unsigned
+                        let res: u64 = if value < 0.0 || value.fract() != 0.0 {
+                            0
+                        } else {
+                            value as u64
+                        };
```
TheElectronWill, line 69: Is there a reason for parsing as `f64` and then only keeping values that are unsigned integers?

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -272,7 +273,8 @@ impl AlumetPlugin for K8sPlugin {
 
 impl Default for K8sConfig {
     fn default() -> Self {
-        let root_path = PathBuf::from("/sys/fs/cgroup/kubepods.slice/");
+        let root_path = PathBuf::from("/sys/fs/cgroup/system.slice/docker-3525f4fbfe549c7eb566acec93e89a57124659a55e0a16531e47ff899cf6bc49.scope/kubepods.slice/");
+        //let root_path = PathBuf::from("/sys/fs/cgroup/kubepods.slice/");
```
TheElectronWill, line 277: This should not have been committed

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -38,13 +39,12 @@ impl Token {
...
-            Err(_) => return false,
-        };
+        let now_secs: u64 = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("ERROR : Failed to get the current time since UNIX_EPOCH")
+            .as_secs();
 
-        // Lock the data for the entire check.
-        let data_locked = self.data.read().await;
+        let data_locked: tokio::sync::RwLockReadGuard<'_, TokenData> = self.data.read().await;
```
TheElectronWill, line 47: ?

TheElectronWill: pro-tip: you can configure VSCode to display the types of the variables on demand.
image

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -55,70 +55,65 @@ impl Token {
...
+                    None => Err(anyhow!("ERROR : Could not read the token's value")),
+                }
+            }
+            false => match self.refresh().await {
+                Ok(token) => Ok(token),
+                Err(e) => Err(anyhow!("ERROR : Failed to refresh token: {}", e)),
+            },
         }
-
-        self.refresh().await
```
TheElectronWill, line 65: How is the new version better than the old one? If all you want is to add a message when `refresh` fails, use `context` or `with_context`: https://docs.rs/anyhow/latest/anyhow/

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -55,70 +55,65 @@ impl Token {
...
     }
 
     /// Retrieves the k8s API token using either a kubectl command
     /// or by reading  the service account token's file.
     async fn refresh(&self) -> anyhow::Result<String> {
-        let token = match self.retrieval {
+        let token: String = match self.retrieval {
             TokenRetrieval::Kubectl => {
-                let output = Command::new("kubectl")
+                let output: std::process::Output = Command::new("kubectl")
```
TheElectronWill, line 78: ?

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -55,70 +55,65 @@ impl Token {
...
+                String::from_utf8(output.stdout)
+                    .context("ERROR : UTF-8 output conversion")?
+                    .trim()
+                    .to_string()
             }
+
             TokenRetrieval::File => {
                 let mut file = File::open("/var/run/secrets/kubernetes.io/serviceaccount/token")
-                    .map_err(|e| anyhow!("Failed to open service account token file: {}", e))?;
+                    .context("ERROR : opening token file")?;
```
TheElectronWill, line 91: yes :)

TheElectronWill: However, don't prefix every error message with `ERROR`. An "ERROR" prefix will be automatically be printed by the log library anyway.
See https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=errors#examples-of-error-messages

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -55,70 +55,65 @@ impl Token {
...
                 let mut token = String::new();
-                file.read_to_string(&mut token)
-                    .map_err(|e| anyhow!("Failed to read service account token file: {}", e))?;
+                file.read_to_string(&mut token).expect("ERROR : Reading token file");
```
TheElectronWill, line 93: You should probably use `context` and `?` here.
Please read https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -55,70 +55,65 @@ impl Token {
...
+            serde_json::from_slice(&decoded_payload).context("ERROR : JSON payload parsing")?;
+
+        let exp: u64 = payload_json
+            .get("exp")
+            .and_then(|exp| exp.as_u64())
+            .context("ERROR : Missing or invalid 'exp' field in token")?;
 
         {
-            let mut new_data = self.data.write().await;
+            let mut new_data: tokio::sync::RwLockWriteGuard<'_, TokenData> = self.data.write().await;
```
TheElectronWill, line 116: Please don't add explicit types in this situation, it's only creating more "diffs" to review.

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -139,54 +134,74 @@ mod tests {
 
     // Test `refresh` function with a kubectl simulated
     #[tokio::test]
-    async fn test_refresh_1() {
-        let _m = mock("POST", "/api/v1/namespaces/alumet/serviceaccounts/alumet-reader/token")
```
TheElectronWill: Why has the API URL changed?

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -139,54 +134,74 @@ mod tests {
...
-        //assert!(data.expiration_time.is_some());
+        assert!(result.is_err());
+        assert_eq!(result.unwrap_err().to_string(), "ERROR : kubectl command execution");
     }
 
     // Test `refresh` function with a file
     #[tokio::test]
-    async fn test_refresh_2() {
+    async fn test_refresh_kubectl_utf8_error() {
+        let _m: mockito::Mock = mock("POST", "/api/alumet/kubernetes.io/serviceaccounts/token")
```
TheElectronWill, line 153: Nice use of mockito

TheElectronWill (changes requested):

plugin-cgroupv2/src/oar3/probe.rs
```diff
@@ -37,14 +42,53 @@ impl CgroupV2prob {
...
+    ///
+    /// # Arguments
+    ///
+    /// - `timestamp` : Type Timestamp
+    /// - `metric_id` : TypedMetricId<u64>
+    /// - `resource_consumer` : Type ResourceConsumer
+    /// - `value_measured` : Type u64
+    /// - `metrics_param` : Type CgroupV2Metric
+    ///
+    fn create_measurement_point(
```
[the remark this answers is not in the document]
TheElectronWill: While we are at it: the `name` attribute could be renamed to `pod_name`, what do you think?

plugin-cgroupv2/src/oar3/plugin.rs
```diff
@@ -188,9 +191,65 @@ impl AlumetPlugin for OARPlugin {
...
+    fn test_default_config() {
+        let result = OARPlugin::default_config().unwrap();
+        assert!(result.is_some(), "Expected Some(ConfigTable): result = None");
+
+        let config_table = result.unwrap();
+        let config: OAR3Config = deserialize_config(config_table).expect("Failed to deserialize config");
+
+        assert_eq!(config.path, PathBuf::from("/sys/fs/cgroup/"));
+        assert_eq!(config.poll_interval, Duration::from_secs(1));
+    }
```
TheElectronWill, line 236: I know that you want to increase the code coverage, but:
- `serialize_config` and `deserialize_config` are already tested elsewhere
- comparing the values with what `Default::default` returns is just duplicating code IMO
- it could be written more cleanly as `assert_eq!(Some(OAR3Config::default()), OARPlugin::default_config().unwrap())` (and deriving Eq for the config).

In the end, I'm not even convinced it's worth spending time to check that :thinking:

TheElectronWill:
You should overwrite your previous commit (with git amend + git push --force-with-lease) when you fix the tests.

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -137,18 +117,51 @@ impl Token {
...
+    }
 
+    // Test `is_valid` function
     #[tokio::test]
     async fn test_is_valid() {
         let token: Token = Token::new(TokenRetrieval::File);
         assert!(!token.is_valid().await);
-
         {
-            // Expand the expiration of the token to make it still valid.
```
titouanj, line 151: ?

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -162,9 +175,7 @@ mod tests {
         }
 
         assert!(token.is_valid().await);
-
         {
-            // Decrease the expiration of the token to make it invalid.
```
titouanj, line 167: ?

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -68,62 +68,42 @@ impl Token {
     /// Retrieves the k8s API token using either a kubectl command
     /// or by reading  the service account token's file.
     async fn refresh(&self) -> anyhow::Result<String> {
-        let token = match self.retrieval {
+        let token: String = match self.retrieval {
```
titouanj, line 71:
```suggestion
        let token = match self.retrieval {
```

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -68,62 +68,42 @@ impl Token {
...
                 let output = Command::new("kubectl")
                     .args(["create", "token", "alumet-reader", "-n", "alumet"])
-                    .output()?;
-
-                if !output.status.success() {
-                    return Err(anyhow!(
-                        "kubectl raised an error: {}",
-                        String::from_utf8_lossy(&output.stderr)
-                    ));
-                }
```
titouanj, line 82: Why removing this check?

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -68,62 +68,42 @@ impl Token {
...
+                    .to_string()
             }
+
             TokenRetrieval::File => {
-                let mut file = File::open("/var/run/secrets/kubernetes.io/serviceaccount/token")?;
+                let mut file =
+                    File::open("/var/run/secrets/kubernetes.io/serviceaccount/token").context("Opening token file")?;
                 let mut token = String::new();
-                file.read_to_string(&mut token)?;
+                file.read_to_string(&mut token).expect("Reading token file");
```
titouanj, line 88: Nope, we don't want a panic here

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -68,62 +68,42 @@ impl Token {
...
                 token
             }
         };
 
         let mut components = token.split('.');
-        let _ = components
-            .next()
-            .ok_or(anyhow!("Could not parse the token, the header is missing"))?;
```
titouanj, line 98: Have JWT lost their heads?

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -68,62 +68,42 @@ impl Token {
...
-                    .trim(),
-            )?,
-        )?;
-        let _ = components
-            .next()
-            .ok_or(anyhow!("Could not parse the token, the signature is missing"))?;
-
-        if components.next().is_some() {
-            return Err(anyhow!("Token has too many components"));
-        }
```
titouanj, line 113: What happened to the signature of JWTs?

TheElectronWill:
> You should overwrite your previous commit (with git amend + git push --force-with-lease) when you fix the tests.

:eyes:

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -38,12 +40,11 @@ impl Token {
...
-        let now_secs = match SystemTime::now().duration_since(UNIX_EPOCH) {
-            Ok(ts) => ts.as_secs(),
-            Err(_) => return false,
-        };
+        let now_secs = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Failed to get the current time since UNIX_EPOCH")
+            .as_secs();
 
-        // Lock the data for the entire check.
```
titouanj, line 46: ?

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -733,30 +733,18 @@ mod tests {
         Ok(())
     }
 
-    // Test `get_existing_pods` function an invalid JSON response from the Kubernetes
+    // Test `get_pod_name` and `get_existing_pods` `pods
```
titouanj, line 736: A unit test should be run over one function only

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -733,30 +733,18 @@ mod tests {
...
 
         let result = get_pod_name(uid, node, kubernetes_api_url, &token).await;
         assert!(result.is_ok());
 
-        let (name, namespace, node) = result.unwrap();
-        assert!(name.is_empty());
-        assert!(namespace.is_empty());
-        assert!(node.is_empty());
+        let result = get_existing_pods(node, kubernetes_api_url, &token).await;
+        assert!(result.is_ok());
```
titouanj, line 748: This shouldn't be done in the same unit test

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -733,30 +733,18 @@ mod tests {
...
-        let uid = "non-existent-uid";
-        let node = "test-node";
+    async fn test_pods() {
+        let uid = "test_uid";
+        let node = "test_node";
         let kubernetes_api_url = "https://127.0.0.1:8080";
         let token = Token::new(TokenRetrieval::Kubectl);
 
         let result = get_pod_name(uid, node, kubernetes_api_url, &token).await;
         assert!(result.is_ok());
```
titouanj, line 745: You should also check the values returned by get_pod_name

plugin-cgroupv2/src/k8s/utils.rs
titouanj, line 748: You should also check the values returned by get_existing_pods

plugin-cgroupv2/src/k8s/token.rs
titouanj, line 1: Your test functions should have better names
For example, test_refresh_2 could be test_refresh_with_missing_exp
It will make it much easier to debug

titouanj (changes requested):

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -62,17 +64,19 @@ impl Token {
...
 
-        self.refresh().await
+        self.refresh("/var/run/secrets/kubernetes.io/serviceaccount/token")
+            .await
     }
 
     /// Retrieves the k8s API token using either a kubectl command
     /// or by reading  the service account token's file.
-    async fn refresh(&self) -> anyhow::Result<String> {
+    async fn refresh(&self, path: &str) -> anyhow::Result<String> {
```
titouanj, line 73: I think it doesn't make any sense to have path as an argument of the function refresh instead of a token's field. You cannot refresh a token with the File retrieval

TheElectronWill:
If something mentioned in a GitHub comment has been resolved, please click on "resolve conversation" so that we know what is done ^^

victoryeagle77:
[ Not done yet ] : Some unit test can't be made for some file functions, because it constraining to import and use the entire Alumet pipeline. Also, it is necessary to develop integration test and for the following functions :
post_pipeline_start in plugin.rs file for k8s and oar3 plugin
start in plugin.rs file for k8s and oar3 plugin
new in cgroupv2.rs file
[ Done ] : Except these functions, all possible unit test for k8s and oar3 files in cgroupv2-plugin were made

TheElectronWill:
> Some unit test can't be made for some file functions, because it constraining to import and use the entire Alumet pipeline. Also, it is necessary to develop integration test and for the following functions :

This will be done in a future PR, this one is only for "simple" unit tests that don't need the core to be modified.

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -141,9 +146,6 @@ impl AlumetPlugin for K8sPlugin {
                     detector: &mut PodDetector,
                     event: Result<Event, notify::Error>,
                 ) -> Result<(), anyhow::Error> {
-                    // The events look like the following
-                    // Handle_Event: Ok(Event { kind: Create(Folder), paths: ["/sys/fs/cgroup/kubepods.slice/kubepods-besteffort.slice/TESTTTTT"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None })
-                    // Handle_Event: Ok(Event { kind: Remove(Folder), paths: ["/sys/fs/cgroup/kubepods.slice/kubepods-besteffort.slice/TESTTTTT"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None })
```
TheElectronWill, line 146: Please put the comments back

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -137,18 +149,50 @@ impl Token {
...
+    }
+
+    // Test `is_valid` function
     #[tokio::test]
     async fn test_is_valid() {
         let token: Token = Token::new(TokenRetrieval::File);
         assert!(!token.is_valid().await);
-
         {
-            // Expand the expiration of the token to make it still valid.
```
TheElectronWill, line 151: Please put useful comments back

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -34,9 +37,29 @@ pub struct CgroupV2MetricFile {
...
 
-/// Check if a specific file is a dir. Used to know if cgroup v2 are used
-pub fn is_accessible_dir(path: &Path) -> bool {
-    path.is_dir()
+/// Check if a specific file is a dir. Used to know if cgroup v2 are used.
+///
+/// # Return
+///
+/// - Boolean error ocurres when path of a file or directory is not available
+/// - Error ocurres when an other kind of issue appears on the path
```
TheElectronWill, line 45:
```suggestion
/// # Return value
///
/// Returns `Ok(true)` if it can be verified that `path` is a directory, and `Ok(false)` if it can be verified that it is not a directory.
/// Returns an error if the path metadata cannot be obtained.
```

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -34,9 +37,29 @@ pub struct CgroupV2MetricFile {
...
+            }
+        }
+        Err(e) => {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                Ok(false) // Path not exist
+            } else {
+                Err(e) // Permissions errors or other
+            }
+        }
+    }
```
TheElectronWill, line 62:
```suggestion
pub fn is_accessible_dir(path: &Path) -> Result<bool, std::io::Error> {
    match std::fs::metadata(path) {
        Ok(metadata) => Ok(metadata.is_dir()),
        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(false),
        Err(e) => Err(e)
    }
}
```

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -1,3 +1,6 @@
+//! # plugin file for k8s module of cgroupv2 plugin
```
TheElectronWill, line 1: ?

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -468,37 +622,494 @@ mod tests {
...
+            .create();
+
+        let uid = "hash1";
+        let kubernetes_api_url = &mockito::server_url();
+        let mut token = Token::new(TokenRetrieval::File);
+
+        token.path =
+            Some("/tmp/test-alumet-plugin-k8s/var_7/run/secrets/kubernetes.io/serviceaccount/token_7".to_string());
+
+        get_pod_name(uid, node, kubernetes_api_url, &token).await.unwrap();
```
TheElectronWill, line 1001: The result of the function should be checked, otherwise the test does not prevent regressions properly.

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -468,37 +622,494 @@ mod tests {
...
+        let root = tmp.join("test-alumet-plugin-k8s/var_8/");
+
+        if root.exists() {
+            std::fs::remove_dir_all(&root).unwrap();
+        }
+
+        let dir = root.join("run/secrets/kubernetes.io/serviceaccount/");
+        std::fs::create_dir_all(&dir).unwrap();
+
+        let content = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwiZXhwIjo0MTAyNDQ0ODAwLCJuYW1lIjoiVDNzdDFuZyBUMGszbiJ9.3vho4u0hx9QobMNbpDPvorWhTHsK9nSg2pZAGKxeVxA";
```
TheElectronWill, line 1018: The token seems to be the same everywhere, put it in a constant.

plugin-cgroupv2/src/oar3/utils.rs
```diff
@@ -200,31 +335,95 @@ mod tests {
...
+    // Test `list_all_file` function with different file system
+    #[test]
+    fn test_list_all_file() {
+        let tmp = tempfile::tempdir().unwrap();
+        let root = tmp.path();
+
+        let non_existent_path = root.join("non_existent");
+        let result = list_all_file(&non_existent_path);
+        assert!(result.is_ok());
+        assert!(result.unwrap().is_empty());
```
TheElectronWill, line 410: These two lines are redundant. Keep only the last one.

TheElectronWill (changes requested):

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -153,23 +168,52 @@ pub fn list_all_k8s_pods_file(
...
+/// - Error if CPU or memory data file are not found.
 pub fn gather_value(file: &mut CgroupV2MetricFile, content_buffer: &mut String) -> anyhow::Result<CgroupV2Metric> {
-    content_buffer.clear(); //Clear before use
-    file.file
+    content_buffer.clear();
+
+    // CPU cgroup data
+    file.file_cpu
+        .read_to_string(content_buffer)
+        .context("Unable to gather cgroup v2 CPU metrics by reading file")?;
```
AngeCyp, line 187: why did you remove the old more explanatory error handling ?

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -468,37 +613,609 @@ mod tests {
...
+                                "name": "pod2",
+                                "namespace": "default",
+                                "uid": "56789"
+                            },
+                            "spec": {
+                                "nodeName": "node2"
+                            }
+                        }
+                    ]
+                })
```
AngeCyp, line 1032: Maybe you can try with different json, adding unecessary section, differents number of item...

plugin-cgroupv2/src/oar3/probe.rs
```diff
@@ -59,42 +89,106 @@ impl alumet::pipeline::Source for CgroupV2prob {
...
             measurements.push(p_sys);
         }
+
+        // Push anonymous used memory measure corresponding to running process and various allocated memory
+        let mem_anon_value = metrics.anon_used_mem;
+        let m_anon = self.create_measurement_point(
+            timestamp,
+            self.anon_used_mem,
+            consumer_memory.clone(),
+            mem_anon_value / 1000,
```
AngeCyp, line 143: as you have kb, you divide by 1000, is there a specific reason to use kb instead of "raw" unit ?

plugin-cgroupv2/src/oar3/utils.rs
```diff
@@ -200,31 +324,95 @@ mod tests {
...
             assert_eq!(time_used_tot, 8335557927);
             assert_eq!(time_used_user_mode, 4728882396);
             assert_eq!(time_used_system_mode, 3606675531);
+            assert_eq!(anon_used_mem, 8335557927);
+            assert_eq!(file_mem, 4728882396);
+            assert_eq!(kernel_mem, 3686400);
+            assert_eq!(pagetables_mem, 0);
         } else {
-            assert!(false);
+            assert!(result.is_err());
```
AngeCyp, line 384: But in this case, if we are in the else, it's because, result is not Ok() so it's an Error so the assert will always be true. 
Isn't it ?

AngeCyp (changes requested):

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -72,7 +76,8 @@ impl Token {
             TokenRetrieval::Kubectl => {
                 let output = Command::new("kubectl")
                     .args(["create", "token", "alumet-reader", "-n", "alumet"])
-                    .output()?;
+                    .output()
+                    .context("Kubectl command execution")?;
```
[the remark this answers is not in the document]
victoryeagle77: We still have "?" after context

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -31,21 +46,29 @@ impl FromStr for CgroupV2Metric {
...
-                        cgroup_struc_to_ret.time_used_tot = parts[1].parse::<u64>()?;
-                    }
-                    "user_usec" => {
-                        cgroup_struc_to_ret.time_used_user_mode = parts[1].parse::<u64>()?;
-                    }
-                    "system_usec" => {
-                        cgroup_struc_to_ret.time_used_system_mode = parts[1].parse::<u64>()?;
+                let value = parts[1]
+                    .parse::<u64>()
+                    .context(format!("ERROR : Parsing of value : {}", parts[1]))?;
```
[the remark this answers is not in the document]
victoryeagle77: We still have "?" after context

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -87,71 +161,123 @@ impl Metrics {
...
+            shmem 2453504
+            file_mapped 72806400
+            ....";
+
+        let result = CgroupV2Metric::from_str(str).unwrap();
+        assert_eq!(result.anon_used_mem, 18446744073709551615);
+        assert_eq!(result.file_mem, 18446744073709551615);
+        assert_eq!(result.kernel_mem, 18446744073709551615);
+        assert_eq!(result.pagetables_mem, 18446744073709551615);
+    }
```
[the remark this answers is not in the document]
victoryeagle77: Ok I replace 18446744073709551615 by true overflow now (2^64 = 18446744073709551616)

plugin-cgroupv2/src/oar3/probe.rs
victoryeagle77: Finally I will replace this by "raw" unit, for a potentially conversion in an other program, if you want to do this.

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -50,24 +59,22 @@ fn list_metric_file_in_dir(
     let entries = fs::read_dir(root_directory_path)?;
     // Let's create a runtime to await async function and fill hashmap
     let rt = tokio::runtime::Builder::new_current_thread().enable_all().build()?;
-    let main_hash_map: HashMap<String, (String, String, String)> =
-        rt.block_on(async { get_existing_pods(hostname, kubernetes_api_url, token).await })?;
+    let main_hash_map = rt.block_on(async { get_existing_pods(hostname, kubernetes_api_url, token).await })?;
```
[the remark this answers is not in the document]
victoryeagle77: They ask me to remove it, apparently according to the Rust convention.

plugin-cgroupv2/src/k8s/probe.rs
```diff
@@ -37,63 +45,150 @@ impl K8SProbe {
...
-        let consumer = ResourceConsumer::ControlGroup {
-            path: (self.cgroup_v2_metric_file.path.to_string_lossy().to_string().into()),
+
+        // CPU ressource consumer for cpu.stat file in cgroup
+        let consumer_cpu = ResourceConsumer::ControlGroup {
+            path: (self.cgroup_v2_metric_file.path_cpu.to_string_lossy().to_string().into()),
         };
+
+        // Memory ressource consumer for memory.stat file in cgroup
+        let consumer_memory = ResourceConsumer::ControlGroup {
```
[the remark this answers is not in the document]
victoryeagle77: I already tried to create it with a single consumer for CPU and memory, but ResourceConsumer needs one path for the resource, and I can't give two different paths for the same consumer.

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -84,8 +89,14 @@ impl Token {
                 let token = String::from_utf8_lossy(&output.stdout);
                 token.trim().to_string()
             }
+
             TokenRetrieval::File => {
-                let mut file = File::open("/var/run/secrets/kubernetes.io/serviceaccount/token")?;
+                let token_path = match &self.path {
+                    Some(path_value) => path_value.clone(),
+                    None => "/var/run/secrets/kubernetes.io/serviceaccount/token".to_string(),
+                };
```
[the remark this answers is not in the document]
victoryeagle77: The error caused by file access is handled after these lines.

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -260,7 +268,10 @@ impl AlumetPlugin for K8sPlugin {
 
 impl Default for K8sConfig {
     fn default() -> Self {
-        let root_path = PathBuf::from("/sys/fs/cgroup/kubepods.slice/");
+        let root_path = PathBuf::from("/sys/fs/cgroup/system.slice/docker-3525f4fbfe549c7eb566acec93e89a57124659a55e0a16531e47ff899cf6bc49.scope/kubepods.slice/");
```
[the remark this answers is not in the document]
victoryeagle77: It was just for test I will remove it immediately.

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -270,3 +281,60 @@ impl Default for K8sConfig {
...
+mod tests {
+    use super::*;
+    use anyhow::Result;
+    use std::path::PathBuf;
+
+    // Create a fake plugin structure for k8s plugin
+    fn fake_k8s() -> K8sPlugin {
+        K8sPlugin {
+            config: K8sConfig {
+                path: PathBuf::from("/sys/fs/cgroup/system.slice/docker-3525f4fbfe549c7eb566acec93e89a57124659a55e0a16531e47ff899cf6bc49.scope/kubepods.slice/"),
```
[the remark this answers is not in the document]
victoryeagle77: Yes but It was just for test I will remove it immediately.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -31,21 +46,29 @@ impl FromStr for CgroupV2Metric {
...
+                        "user_usec" => cgroup_struc_to_ret.time_used_user_mode = value,
+                        "system_usec" => cgroup_struc_to_ret.time_used_system_mode = value,
+                        "anon" => cgroup_struc_to_ret.anon_used_mem = value,
+                        "file" => cgroup_struc_to_ret.file_mem = value,
+                        "kernel_stack" => cgroup_struc_to_ret.kernel_mem = value,
+                        "pagetables" => cgroup_struc_to_ret.pagetables_mem = value,
+                        _ => continue,
                     }
-                    &_ => continue,
                 }
```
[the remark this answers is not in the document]
TheElectronWill: Why have you marked this as "resolved"? The brackets are still there and you have not explained why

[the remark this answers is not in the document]
TheElectronWill: Why have you marked this as "resolved"? The brackets are still there and you have not explained why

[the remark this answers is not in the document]
TheElectronWill: Why have you marked this as "resolved"? The brackets are still there and you have not explained why...

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -1,22 +1,37 @@
...
     plugin::AlumetPluginStart,
     units::{PrefixedUnit, Unit},
 };
+use anyhow::{Context, Result};
+use std::str::FromStr;
 
 pub(crate) const CGROUP_MAX_TIME_COUNTER: u64 = u64::MAX;
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct CgroupV2Metric {
```
TheElectronWill, line 12: Since tests can be boring, forgive me for giving you modifications beyond that :)

`CgroupV2Metric` is not a metric and does not contain metrics. Rename it to `CgroupMeasurements` or something like that.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -1,22 +1,37 @@
...
+    /// Name of a Kubernetes pod.
     pub name: String,
```
TheElectronWill, line 14: Rename `name` to `pod_name`.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -1,22 +1,37 @@
...
+    /// Unique identification of a Kubernetes pod.
     pub uid: String,
```
TheElectronWill, line 16: Rename `uid` to `pod_uid`. Same for all pod-related values.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -1,22 +1,37 @@
...
+    /// Resources isolation of a Kubernetes pod.
     pub namespace: String,
+    /// Kubernetes pod node.
     pub node: String,
+    /// Total CPU usage time by the cgroup.
     pub time_used_tot: u64,
```
TheElectronWill, line 22: Rename to `cpu_time_total`. Same for `time_used_<x>_mode` -> `cpu_time_<x>`

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -1,22 +1,37 @@
...
+    /// CPU in user mode usage time by the cgroup.
     pub time_used_user_mode: u64,
+    /// CPU in system mode usage time by the cgroup.
     pub time_used_system_mode: u64,
+    /// Anonymous used memory, corresponding to running process and various allocated memory.
+    pub anon_used_mem: u64,
```
TheElectronWill, line 28: Rename to: `memory_anonymous`. Same for other memory values: put "memory" first (so it's easier to distinguish between cpu time and memory fields).

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -31,21 +46,27 @@ impl FromStr for CgroupV2Metric {
...
+            kernel_mem: 0,
+            pagetables_mem: 0,
         };
+
         for line in s.lines() {
             let parts: Vec<&str> = line.split_ascii_whitespace().collect();
             if parts.len() >= 2 {
+                let value = parts[1]
+                    .parse::<u64>()
+                    .context(format!("ERROR : Parsing of value : {}", parts[1]))?;
```
TheElectronWill, line 60: Use `with_context` instead of `context`.
Difference:
- `context` always generate the error message even if there's no error
`with_context` does nothing if there's no error -> more efficient

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -55,15 +76,43 @@ impl FromStr for CgroupV2Metric {
 
 #[derive(Clone)]
 pub struct Metrics {
```
TheElectronWill, line 78: Here the name of the structure is okay :)
Apply my remark about the renaming here too.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -55,15 +76,43 @@ impl FromStr for CgroupV2Metric {
...
+    /// Provides a information base to create metric before sending CPU and memory data,
+    /// with `name`, `unit` and `description` parameters.
+    ///
+    /// # Arguments
+    ///
+    /// * `alumet` - A AlumetPluginStart structure passed to plugins for the start-up phase.
+    ///
+    /// # Returns
+    ///
+    /// * `Result` - Type representing success `Ok` or failure `Err`.
```
TheElectronWill, line 107: Don't document what `Result` is: it's basic Rust.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -55,15 +76,43 @@ impl FromStr for CgroupV2Metric {
...
+    /// * `MetricCreationError` - Error which can occur when creating a new metric.
```
TheElectronWill, line 108: Instead, just write that it returns an error if the metric creation fails.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -55,15 +76,43 @@ impl FromStr for CgroupV2Metric {
...
+    ///
     pub fn new(alumet: &mut AlumetPluginStart) -> Result<Self, MetricCreationError> {
-        let usec: PrefixedUnit = PrefixedUnit::micro(Unit::Second);
+        let usec = PrefixedUnit::micro(Unit::Second);
+        let raw = Unit::Byte;
```
TheElectronWill, line 112: Why `raw`? With a simple unit like that, don't even bother creating a local variable and cloning it, use `Unit::Byte` directly.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -87,71 +163,176 @@ impl Metrics {
 mod tests {
     use super::*;
 
+    // Test `from_str` function with in extracted result,
```
TheElectronWill, line 166: Comment is outdated, please update.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -87,71 +163,176 @@ impl Metrics {
...
-        system_usec 33\n
-        user_usec 222222222222222222\n
-        throttled_usec 0"
-        );
-        let result: CgroupV2Metric = CgroupV2Metric::from_str(&str4).unwrap();
-        assert_eq!(result.name, "");
-        assert_eq!(result.time_used_tot, 1111 as u64);
-        assert_eq!(result.time_used_user_mode, 222222222222222222 as u64);
-        assert_eq!(result.time_used_system_mode, 33 as u64);
+    fn test_zero_values() {
```
TheElectronWill, line 169: Ce test (et ceux en dessous) testent en fait `str::parse<u64>` qui est dans la lib standard, donc déjà testé, donc c'est inutile.

Ce qu'il faut tester c'est 
- Que se passe-t-il s'il manque une ligne ? (vérifie la doc noyau pour voir si les champs sont optionnels selon la version du noyau ou non) Si un champ est optionnel, il faut distinguer le cas "pas là" du cas "là mais zéro", par exemple en utilisant `Option`.
- Que se passe-t-il s'il y a d'autres lignes qui ne nous intéressent pas ?
- Que faire si le fichier n'est pas au format attendu ? Si le format est fixé par le noyau, on peut argumenter que c'est pas grave de paniquer dans ce cas (noyau cassé = tout est cassé, donc pas notre problème) mais il faut y réfléchir et documenter le choix.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -87,71 +163,176 @@ impl Metrics {
...
+            file_mapped 0
+            ....";
+
+        let result_cpu = CgroupV2Metric::from_str(str_cpu).unwrap();
+        assert_eq!(result_cpu.time_used_tot, 0);
+        assert_eq!(result_cpu.time_used_user_mode, 0);
+        assert_eq!(result_cpu.time_used_system_mode, 0);
+
+        let result_memory = CgroupV2Metric::from_str(str_memory).unwrap();
+        assert_eq!(result_memory.anon_used_mem, 0);
```
TheElectronWill, line 195: Pro-tip : dérive Eq et PartialEq pour la structure de mesure et tu pourras faire `assert_eq(result, LaStructure { tousleschampsici }`

La dérivation se fait comme ça :
```rs
#[derive(Eq, PartialEq)]
struct LaStructure {
```

Si besoin tu peux dériver Debug aussi.

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -73,8 +72,12 @@ impl AlumetPlugin for K8sPlugin {
         }))
     }
 
+    fn stop(&mut self) -> anyhow::Result<()> {
+        Ok(())
+    }
+
     fn start(&mut self, alumet: &mut AlumetPluginStart) -> anyhow::Result<()> {
-        let v2_used: bool = super::utils::is_accessible_dir(&PathBuf::from("/sys/fs/cgroup/"));
+        let v2_used = super::utils::is_accessible_dir(&PathBuf::from("/sys/fs/cgroup/"))?;
```
TheElectronWill, line 80: Est-ce que `is_accessible_dir` affiche le chemin dans le message d'erreur?

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -73,8 +72,12 @@ impl AlumetPlugin for K8sPlugin {
...
         if !v2_used {
```
TheElectronWill, line 81: Pas convaincu par le `bail!` (ping Cyprien), que penses-tu d'un `return Err`?

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -270,3 +284,60 @@ impl Default for K8sConfig {
...
+                token_retrieval: TokenRetrieval::Kubectl,
+            },
+            watcher: None,
+            metrics: None,
+        }
+    }
+
+    // Test `default_config` function of k8s plugin
+    #[test]
+    fn test_default_config() {
```
TheElectronWill, line 311: On teste quoi exactement ici ? On connaît la default config du plugin, c'est nous qui l'avons écrite. Pourquoi écrire 2 fois la même config et vérifier si c'est égal ? (encore une fois le but principal c'est de s'assurer que le plugin marche, pas d'avoir 100% de couverture. La couverture c'est un indicateur)

plugin-cgroupv2/src/k8s/plugin.rs
```diff
@@ -270,3 +284,60 @@ impl Default for K8sConfig {
...
+        let plugin = K8sPlugin::init(config_table)?;
+        assert_eq!(plugin.config.kubernetes_api_url, "https://127.0.0.1:8080");
+        assert!(plugin.metrics.is_none());
+        assert!(plugin.watcher.is_none());
+        Ok(())
+    }
+
+    // Test `stop` function to stop k8s plugin
+    #[test]
+    fn test_stop() {
```
TheElectronWill, line 338: Tester stop comme ça n'a pas trop de sens de mon point de vue: si stop a des choses à faire, ce sera forcément en lien avec l'utilisation du plugin dans un pipeline Alumet qui n'existe pas ici.

plugin-cgroupv2/src/k8s/probe.rs
```diff
@@ -37,63 +45,150 @@ impl K8SProbe {
...
+            path: (self
+                .cgroup_v2_metric_file
+                .path_memory
+                .to_string_lossy()
+                .to_string()
```
TheElectronWill, line 102: Je pense qu'on peut trouver plus court que `.to_string_lossy().to_string().into()`

plugin-cgroupv2/src/k8s/probe.rs
```diff
@@ -37,63 +45,150 @@ impl K8SProbe {
...
             measurements.push(p_sys);
         }
+
+        // Push anonymous used memory measure corresponding to running process and various allocated memory
+        let mem_anon_value = metrics.anon_used_mem;
+        let m_anon = self.create_measurement_point(
+            timestamp,
+            self.anon_used_mem,
+            consumer_memory.clone(),
+            mem_anon_value,
```
TheElectronWill, line 143: Utilise directement `metrics.xyz` ici, pas besoin de variable intermédiaire.

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -6,18 +6,21 @@ use std::{
...
 use base64::{prelude::BASE64_STANDARD_NO_PAD, Engine};
 use tokio::sync::RwLock;
 
 use super::plugin::TokenRetrieval;
 
+#[derive(Debug)]
 pub struct Token {
     retrieval: TokenRetrieval,
     data: Arc<RwLock<TokenData>>,
+    pub path: Option<String>,
```
TheElectronWill, line 19: Petite doc ?

TheElectronWill (changes requested):
Partie 1

plugin-cgroupv2/src/k8s/plugin.rs
victoryeagle77: Non mais maintenant c'est ajouté

plugin-cgroupv2/src/oar3/probe.rs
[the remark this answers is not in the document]
TheElectronWill: à déplacer dans poll

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -34,9 +34,18 @@ pub struct CgroupV2MetricFile {
...
-/// Check if a specific file is a dir. Used to know if cgroup v2 are used
-pub fn is_accessible_dir(path: &Path) -> bool {
-    path.is_dir()
+/// Check if a specific file is a dir. Used to know if cgroup v2 are used.
+///
+/// # Return value
+///
+/// Returns `Ok(true)` if it can be verified that `path` is a directory, and `Ok(false)` if it can be verified that it is not a directory.
+/// Returns an error if the path metadata cannot be obtained.
+pub fn is_accessible_dir(path: &Path) -> Result<bool, std::io::Error> {
```
TheElectronWill, line 43: Cette fonction (et peut-être d'autres) est dupliquée dans `k8s/utils.rs` et `oar3/utils.rs`. Ce serait bien de dédupliquer le code avec un `src/utils.rs` par exemple, ou même dans `src/lib.rs` si c'est petit.

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -135,7 +150,7 @@ pub fn list_all_k8s_pods_file(
             .file_name()
             .ok_or_else(|| anyhow::anyhow!("No file name found"))?
             .to_str()
-            .with_context(|| format!("Filename is not valid UTF-8: {:?}", path))?;
+            .context("Filename is not valid UTF-8")?;
```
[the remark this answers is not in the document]
TheElectronWill: il faut laisser le chemin vers le fichier dans l'erreur

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -377,46 +437,79 @@ pub async fn get_pod_name(
...
-        let cgroupv2_dir = root.join("myDirCgroup");
-        std::fs::create_dir_all(&cgroupv2_dir).unwrap();
-        assert!(is_accessible_dir(Path::new(&cgroupv2_dir)));
-        assert!(!is_accessible_dir(std::path::Path::new(
-            "test-alumet-plugin-k8s/is_cgroupv2/myDirCgroup_bad"
-        )));
+
+        let dir = root.join("dir_cgroup");
+        std::fs::create_dir_all(&dir).unwrap();
+        assert!(is_accessible_dir(&dir).unwrap());
```
TheElectronWill, line 462:
```suggestion
        assert_eq!(is_accessible_dir(&dir), Ok(true));
```

TheElectronWill: idem ailleurs

TheElectronWill (changes requested):
Pour les dossiers de tests, utilise `tempfile::tempdir()` partout au lieu de construire un truc à la main avec `std::env::temp_dir()`. Sinon on risque d'avoir le même dossier temporaire pour plusieurs tests et ça va poser problème (c'est facile d'oublier de changer le nom quand il y a autant de tests).

plugin-cgroupv2/src/k8s/utils.rs
victoryeagle77: binary operation `==` cannot be applied to type `std::result::Result<bool, std::io::Error>`

plugin-cgroupv2/src/k8s/probe.rs
victoryeagle77: Sachant que je calcule plus loin la mémoire totale avec ces valeurs, je ne peux pas me passer de leur déclaration.

plugin-cgroupv2/src/k8s/probe.rs
victoryeagle77: Ceci a été résolu avec l'utilisation des ResourceConsumer CPU et Memory à la place de leur path.

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -435,30 +528,82 @@ mod tests {
...
+            file_memory,
+            uid: "test-uid".to_string(),
+            namespace: "default".to_string(),
+            node: "test-node".to_string(),
+        };
+
+        let mut content_buffer = String::new();
+        let result = gather_value(&mut metric_file, &mut content_buffer);
+
+        result.expect("gather_value get invalid data");
```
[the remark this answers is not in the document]
victoryeagle77: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `FromResidual`)
the trait `FromResidual<std::result::Result<Infallible, anyhow::Error>>` is not implemented for `()`

plugin-cgroupv2/src/cgroupv2.rs
victoryeagle77: * Que se passe-t-il s'il manque une ligne ? -> Ce test fait avec des lignes en moins et en trop dans le fichier memory.stat

plugin-cgroupv2/src/k8s/plugin.rs
victoryeagle77: Pour l'instant je gardes ces tests car c'est les seuls que je peux faire pour plugin.rs, et je les enleverais et remplacerais lorsque je pourrais faire des tests d'intégration.

plugin-cgroupv2/src/k8s/plugin.rs
victoryeagle77: Pour l'instant je gardes ces tests car c'est les seuls que je peux faire pour plugin.rs, et je les enleverais et remplacerais lorsque je pourrais faire des tests d'intégration.

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -377,46 +437,54 @@ pub async fn get_pod_name(
...
-        std::fs::create_dir_all(&cgroupv2_dir).unwrap();
-        assert!(is_accessible_dir(Path::new(&cgroupv2_dir)));
-        assert!(!is_accessible_dir(std::path::Path::new(
-            "test-alumet-plugin-k8s/is_cgroupv2/myDirCgroup_bad"
-        )));
-    }
+    // HEADER = { "alg": "HS256", "typ": "JWT" }
+    // PAYLOAD = { "sub": "1234567890", "exp": 4102444800, "name": "T3st1ng T0k3n" }
+    // SIGNATURE = { HMACSHA256(base64UrlEncode(header) + "." +  base64UrlEncode(payload), signature) }
+    const TOKEN_CONTENT: &str = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwiZXhwIjo0MTAyNDQ0ODAwLCJuYW1lIjoiVDNzdDFuZyBUMGszbiJ9.3vho4u0hx9QobMNbpDPvorWhTHsK9nSg2pZAGKxeVxA";
```
AngeCyp, line 448:
```suggestion
    // cspell:disable-next-line
    const TOKEN_CONTENT: &str = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwiZXhwIjo0MTAyNDQ0ODAwLCJuYW1lIjoiVDNzdDFuZyBUMGszbiJ9.3vho4u0hx9QobMNbpDPvorWhTHsK9nSg2pZAGKxeVxA";
```

AngeCyp (changes requested):
Don't hesitate to do:
`
cargo fmt
`
To resolve lint issues of the CI

plugin-cgroupv2/src/k8s/utils.rs
AngeCyp: Do you agree @titouanj  If yes, @victoryeagle77  you can do it for every cpsell error like it

TheElectronWill:
For the doc:
```rs
// spell-checker: disable
le token ici
// spell-checker: enable

File: plugin-cgroupv2/src/k8s/utils.rs (hunk 81 above)
Comment (TheElectronWill) [reply plugin-cgroupv2/src/k8s/utils.rs:448]: yes

File: plugin-cgroupv2/Cargo.toml [hunk 82]
```diff
@@ -17,6 +17,8 @@ serde_json = "1.0"
 humantime-serde = "1.1.1"
 notify = "6.1.1"
 base64 = "0.22.1"
+tempfile = "3.5"
+mockito = "0.31"
```
TheElectronWill, line 21: Move these to development-only dependencies, as they are only used in tests.
```toml
[dev-dependencies]
tempfile = "3.5"
mockito = "0.31"
```

plugin-cgroupv2/src/k8s/plugin.rs
TheElectronWill: fair enough :)

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -377,46 +437,56 @@ pub async fn get_pod_name(
...
+    // spell-checker: enable
 
+    // Test `list_metric_file_in_dir` function to simulate arborescence of kubernetes pods
+    // and missing files in Kubernetes directory
     #[test]
     fn test_list_metric_file_in_dir() {
         let tmp = std::env::temp_dir();
```
TheElectronWill, line 456: Use tempfile::tempdir() everywhere

TheElectronWill (changes requested):

plugin-cgroupv2/src/k8s/utils.rs
```diff
@@ -468,37 +607,736 @@ mod tests {
...
+                })
+                .to_string(),
+            )
+            .create();
+
+        let kubernetes_api_url = &mockito::server_url();
+        let mut token = Token::new(TokenRetrieval::File);
+
+        token.path =
+            Some("/tmp/test-alumet-plugin-k8s/var_5/run/secrets/kubernetes.io/serviceaccount/token_5".to_string());
```
TheElectronWill, line 894: Don't hardcode the path, and use `tempfile::tempdir` to get a new unique temporary directory.

TheElectronWill (changes requested):

plugin-cgroupv2/src/k8s/utils.rs
victoryeagle77: Ah oui je l'ai avais pas tous vu normalement maintenant c'est fait

.cspell/project-words.txt
```diff
@@ -23,6 +23,7 @@ miri
 monitorable
 mpoint
 NVML
+pagetables
```
TheElectronWill, line 26: hot take: `pagetables`, `mpoint`, and others should not be in the dictionary
I am now wondering whether it's a good idea to spell-check every string...

TheElectronWill: (this does not need to be resolved now)

plugin-cgroupv2/src/lib.rs
```diff
@@ -2,5 +2,63 @@ mod cgroupv2;
...
+#[cfg(test)]
+mod tests {
+    use crate::is_accessible_dir;
+    use std::path::{Path, PathBuf};
+
+    // Tests `is_accessible_dir` function to check the existence of cgroupv2 file system
+    #[test]
+    fn test_is_accessible_dir() {
+        let tmp = std::env::temp_dir();
+        let root = tmp.join("test-alumet-plugin-k8s/kubepods-list-metrics");
```
TheElectronWill, line 33: use tempfile::tempdir

TheElectronWill (changes requested):

plugin-cgroupv2/src/lib.rs
TheElectronWill: last remark :)

victoryeagle77: je l'avais pas vu non plus celui-là

TheElectronWill (approved):

PR #68, opened 2024-11-25
Feature/mongodb plugin

kelyaenn:
This plugin will (maybe) ensure smooth export to a mongodb instance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
     "plugin-socket-control",
     "test-dynamic-plugin-rust",
     "test-dynamic-plugins",
+    "plugin-mongodb",
 ]
 
 # To apply the following lint settings to a workspace member,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,9 @@ members = [
     "plugin-rapl",
     "plugin-relay",
     "plugin-socket-control",
+    "plugin-mongodb",
     "test-dynamic-plugin-rust",
     "test-dynamic-plugins",
-    "plugin-mongodb",
 ]
 
 # To apply the following lint settings to a workspace member,
--- a/plugin-mongodb/Cargo.toml
+++ b/plugin-mongodb/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "plugin-mongodb"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+alumet = { path = "../alumet" }
+anyhow = "1.0.93"
+log = "0.4.22"
+serde = "1.0.215"
+tokio = "1.41.1"
+
+[dependencies.mongodb]
+version = "3.1.0"
+features = ["sync"]
+
+
+[lints]
+workspace = true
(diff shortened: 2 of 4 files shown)


plugin-mongodb/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+            // Add the timestamp
+            doc.insert("timestamp", convert_timestamp(m.timestamp));
+
+            docs.push(doc);
+        }
+
+        // Send the data to MongoDb
+        let collection = self.client.database(&self.database).collection(&self.collection);
+        let rt = tokio::runtime::Runtime::new().unwrap();
```
TheElectronWill, line 148: You should not create a new normal runtime on each `write`, because `write` is already called in a runtime by Alumet. Either you use the current-thread runtime, or you turn your output into an "async" output (See `AlumetPluginStart::add_async_output_builder`).

plugin-mongodb/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+impl Output for MongoDbOutput {
+    fn write(&mut self, measurements: &MeasurementBuffer, ctx: &OutputContext) -> Result<(), WriteError> {
+        // Cannot write anything with an empty buffer.
+        if measurements.is_empty() {
+            log::warn!("MongoDb recieved an empty MeasurementBuffer");
+            return Ok(());
+        }
+
+        // Build the data to send to MongoDb
+        let mut docs = vec![];
```
TheElectronWill, line 81:
```suggestion
        let mut docs = Vec::with_capacity(measurements.len());
```

plugin-mongodb/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[dependencies]
+alumet = { path = "../alumet" }
+anyhow = "1.0.93"
+log = "0.4.22"
+serde = "1.0.215"
+tokio = "1.41.1"
+
+[dependencies.mongodb]
+version = "3.1.0"
+features = ["sync"]
```
TheElectronWill, line 15:
```suggestion
mongodb = { version = "3.1.0", features = ["sync"] }
```
it's easier to read this way (and more coherent with the whole ecosystem)

plugin-mongodb/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+        for m in measurements {
+            let mut doc = doc! {};
```
TheElectronWill, line 83:
```suggestion
            let mut doc = Document::new();
```
personal preference (minor nitpick :D): the document is empty here, so the macro does the same as `Document::new()` but obscures the type and adds (compilation) overhead.

Cargo.toml
```diff
@@ -18,6 +18,7 @@ members = [
     "plugin-socket-control",
     "test-dynamic-plugin-rust",
     "test-dynamic-plugins",
+    "plugin-mongodb",
```
TheElectronWill, line 21: Could you move it above the test crates?

TheElectronWill (changes requested):
Thanks for the new plugin! Overall it's good, the main remark is about the use of tokio Runtimes.

plugin-mongodb/Cargo.toml
kelyaenn: I didn't even know this was possible.
I'm still a Rust newbie 😃

PR #71, opened 2024-11-27
agent: Improve exec error message

### LINKED ISSUE #6: Improve error message for exec mode
## How to reproduce
Try to spawn and monitor a particular program `my_app.py` that exists in the current directory:
```sh
alumet-agent exec my_app.py
```
Error:
```
thread 'main' panicked at app-agent/src/main.rs:81:18:
error in child process: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Alumet is right, the correct command is:
```sh
alumet-agent exec ./my_app.py
```
but the error message should be improved. Also, could it detect such a case and suggest the right command?

AngeCyp:
Fix #6 

This PR add a better message when using the exec mode and if the executable is not found.


app-agent/src/exec_process.rs
```diff
@@ -17,10 +17,73 @@ use anyhow::Context;
...
+    let p_result = std::process::Command::new(external_command.clone())
+        .args(args.clone())
+        .spawn();
+
+    let mut p = match p_result {
+        Ok(val) => val,
+        Err(e) => match e.kind() {
+            std::io::ErrorKind::NotFound => {
+                let directory_entries_iter = fs::read_dir("./").unwrap();
+                let mut entry_list = String::from("Available files in the current directory:\n");
```
TheElectronWill, line 29: I don't think it's up to the agent to replace `ls`.
How about just doing these two things:

## If `ErrorKind::NotFound`

- check if a file with the same name exists in the current directory
- print:
```
command 'XYZ' not found
Hint: A file named 'XYZ' exists in the current directory. Prepend ./ to execute it.
Example: CMD exec './XYZ'
```
where `CMD` is the alumet binary (in the rapl plugin you can find an example of how to get the path to the current alumet binary on the user).

### Bonus

If you want to have fun, implement a fuzzy finder to fix the potential typos :)
Example of what it would look like:
```sh
> ls
script.sh

> alumet-local-agent exec ./scrit.sh
ERROR: ...
Hint: did you mean './script.sh'?
```

Other example:
```sh
> ls
script.sh

> alumet-local-agent exec ./script
ERROR: ...
Hint: did you mean './script.sh'?
```

## If permission denied
- check if the file exists but does not have the X or R permission
- print:
```
file 'XYZ' is missing the following permissions: PERM
Hint: try chmod +PERM 'XYZ'
```
where `PERM` is `r`, `x` or `rx`

TheElectronWill (changes requested):

.cspell/project-words.txt
```diff
@@ -2,6 +2,7 @@ alumet
 ASCIINEMA
 astring
 Atos
+capaldi
```
TheElectronWill, line 5: Peter? :D

app-agent/src/exec_process.rs
```diff
@@ -17,10 +22,26 @@ use anyhow::Context;
...
+                let return_error: String = handle_not_found(external_command, args);
+                panic!("{}", return_error);
+            }
+            std::io::ErrorKind::PermissionDenied => {
+                let return_error: String = handle_permission_denied(external_command);
+                panic!("{}", return_error);
+            }
+            _ => {
+                panic!("Error in child process");
+            }
```
TheElectronWill, line 42: The function returns a `Result`, you should use that instead of panics.

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,111 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
     Ok(status)
 }
 
+fn handle_permission_denied(external_command: String) -> String {
+    let file_permission_denied = match File::open(external_command.clone()) {
+        Ok(file) => file,
+        Err(err) => {
+            panic!("Error when try to read the file: {}", err);
```
TheElectronWill, line 61: english

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,111 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+        _ => "",
+    };
+    // Check for other permissions.
+    let other_perm = match metadata_file.permissions().mode() & 0o005 {
+        0 => "rx",
+        1 => "r",
+        4 => "x",
+        _ => "",
+    };
+    if user_perm == "rx" || group_perm == "rx" || other_perm == "rx" {
```
TheElectronWill, line 88: What is all this doing, exactly? Isn't `chmod +rx` a solution that always work?

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,111 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+        }
+    }
+    match best_element {
+        Some((element, distance)) => {
+            if distance == 0 {
+                log::info!(
+                    "💡 Hint: A file named '{}' exists in the current directory. Prepend ./ to execute it.",
+                    element
+                );
+                log::info!("Example: {} exec ./{} {}", app_path, element, args.join(" "));
```
TheElectronWill, line 150: Arguments may need to be quoted and escaped (e.g. if they contain a whitespace)

app-agent/src/utils.rs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            let operations = [
+                d[i][j] + cost,                          // substitution
+                d[i + 1][j] + 1,                         // insertion
+                d[i][j + 1] + 1,                         // deletion
+                d[k][l] + (i - k - 1) + 1 + (j - l - 1), // transposition
+            ];
+            d[i + 1][j + 1] = *operations.iter().min().unwrap();
+        }
+        da.insert(a.chars().nth(i - 1).unwrap(), i);
```
TheElectronWill, line 43: Using `.chars().nth(...)` in a loop is suboptimal because it traverses the string multiple times. Turn the string into a char vec once, before the loop, or do the loop on the `chars()`.

app-agent/src/utils.rs
```diff
@@ -0,0 +1,168 @@
+use std::collections::HashMap;
+
+// Damerau-Levenshtein distance
+pub fn distance_with_adjacent_transposition(a: String, b: String) -> usize {
+    let a_len = a.chars().count();
+    let b_len = b.chars().count();
```
TheElectronWill, line 6: See my comment below: you are traversing the strings multiple times.

TheElectronWill (changes requested):
When done, it will be a nice improvement to the agent `exec` mode!

.cspell/project-words.txt
AngeCyp: Yes, himself

app-agent/src/exec_process.rs
AngeCyp: As in the main you used `expect()` I expected to panic in case of error

app-agent/src/exec_process.rs
AngeCyp: I check which rights are missing, the goal is to suggest only to add missing rights, if read is missing but not execute, i will suggest `chmod +r` only

TheElectronWill: That's very generous of your part, but `chmod` does the checking, no? If `chmod +r` works, `chmod +rx` will work too. Something that would really be helpful is, maybe, to check the permissions of the folder: what happens if the user wants to execute `/a/b/exec` but `/a/b` is not readable?

AngeCyp: Of course, if `+r` is missing, `+rx` will work too. I found it was more interesting for people to see that, for example, if only read right is missing, but the execute one is already present... But it's as you want, I can just suggest one chmod: `chmod +rx`.
About the parent folder, you're totally right, I will check this right in my next commit. I will only check the parent folder but not the parent of the parents, as it could be veryyyyyy long and not necessary, in my opinion

app-agent/src/exec_process.rs
```diff
@@ -12,15 +17,31 @@ use alumet::{
...
+            std::io::ErrorKind::NotFound => {
+                let return_error: String = handle_not_found(external_command, args);
+                return Err(anyhow!(return_error));
+            }
+            std::io::ErrorKind::PermissionDenied => {
+                let return_error: String = handle_permission_denied(external_command);
+                return Err(anyhow!(return_error));
+            }
+            _ => {
+                panic!("Error in child process");
```
TheElectronWill, line 41: `exec_child` returns a result, please remove the panics

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
     Ok(status)
 }
 
+fn handle_permission_denied(external_command: String) -> String {
+    let file_permission_denied = match File::open(external_command.clone()) {
+        Ok(file) => file,
+        Err(err) => {
```
TheElectronWill, line 60: the `Ok` branch is very small, you can turn this into a `if let` instead. Also, what does `file_permission_denied` represent?

It would be cleaner to make it a separate function.

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+                    Err(_) => {
+                        current_parent = match current_parent.parent() {
+                            Some(parent) => parent,
+                            None => {
+                                panic!("Unable to retrieve a parent for your file");
+                            }
+                        }
+                    }
+                }
+            }
```
TheElectronWill, line 84: It would be cleaner to make it a separate function

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+    let file_metadata = file_permission_denied
+        .metadata()
+        .expect(format!("Unable to retrieve metadata for: {}", external_command).as_str());
+    // Check for user permissions.
+    let user_perm = match file_metadata.permissions().mode() & 0o500 {
+        0 => "rx",
+        1 => "r",
+        4 => "x",
+        _ => "",
+    };
```
TheElectronWill, line 133: A function would be handy for deduplication here.

app-agent/src/exec_process.rs
TheElectronWill: If you want to do that, do it properly: compute the missing permissions and print the message that suggests it in one go, without repeating all the possibilities (lines 148 to 162 should be just 2 lines).

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+    } else if user_perm == "r" || group_perm == "r" || other_perm == "r" {
+        log::error!("file '{}' is missing the following permissions:  'r'", external_command);
+        log::info!("💡 Hint: try 'chmod +r {}'", external_command)
+    } else if user_perm == "x" || group_perm == "x" || other_perm == "x" {
+        log::error!("file '{}' is missing the following permissions:  'x'", external_command);
+        log::info!("💡 Hint: try 'chmod +x {}'", external_command)
+    } else {
+        log::warn!("Can't determine right issue about the file: {}", external_command);
+    }
+    "Issue happened about file's permission".to_string()
```
TheElectronWill, line 163: which issue? Use proper errors propagation

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+                    args.iter()
+                        .map(|arg| {
+                            if arg.contains(' ') {
+                                format!("\"{}\"", arg)
+                            } else {
+                                arg.to_string()
+                            }
+                        })
+                        .collect::<Vec<_>>()
+                        .join(" ")
```
TheElectronWill, line 223: Please separate the construction of the list from the log call (same everywhere).

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+                );
+            }
+        }
+        None => {
+            log::warn!("💡 Hint: No matching file exists in the current directory. Prepend ./ to execute it.");
```
TheElectronWill, line 243: What? If there's no matching file, prepending `./` will never work.

app-agent/src/exec_process.rs
```diff
@@ -33,6 +54,198 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+        }
+    }
+    "Issue happened because the file was not found".to_string()
```
TheElectronWill, line 246: which issue?

app-agent/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod agent_util;
 pub mod config_ops;
 pub mod exec_process;
 pub mod options;
+pub mod utils;
```
TheElectronWill, line 9: Please find a better name than `utils`. I suggest `word_distance`

TheElectronWill (changes requested):

app-agent/src/utils.rs
```diff
@@ -0,0 +1,170 @@
+use std::collections::HashMap;
+
+// Damerau-Levenshtein distance
```
TheElectronWill, line 3: Use triple slash to document function. Please add more details so that we understand what this is without googling.

app-agent/src/exec_process.rs
```diff
@@ -17,7 +17,7 @@ use alumet::{
     plugin::event::StartConsumerMeasurement,
     resources::ResourceConsumer,
 };
-use anyhow::{anyhow, Context};
+use anyhow::{anyhow, Context, Error};
 
 /// Spawns a child process and waits for it to exit.
 pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result<ExitStatus> {
```
TheElectronWill, line 23: Please include the error `e` in the returned errors, so that we don't lose the original information.

app-agent/src/exec_process.rs
```diff
@@ -202,48 +138,88 @@ fn handle_not_found(external_command: String, args: Vec<String>) -> String {
...
         }
         None => {
-            log::warn!("💡 Hint: No matching file exists in the current directory. Prepend ./ to execute it.");
+            log::warn!("💡 Hint: No matching file exists in the current directory. Please try again we a different one.");
+        }
+    }
+    "Sorry but the file was not found".to_string()
+}
+
+fn check_missing_permissions(current_permissions: u32, required_permissions: u32) -> u32 {
```
TheElectronWill, line 174: Is the returned number actually used as a number or only as a "rwx" string? If the latter, that string could be constructed directly in this function.

app-agent/src/word_distance.rs
```diff
@@ -1,7 +1,38 @@
...
+/// let distance = distance_with_adjacent_transposition("kitten".to_string(), "sitting".to_string());
+/// assert_eq!(distance, 3);
+/// 
+/// let distance = distance_with_adjacent_transposition("flaw".to_string(), "lawn".to_string());
+/// assert_eq!(distance, 2);
+/// 
+/// let distance = distance_with_adjacent_transposition("ca".to_string(), "abc".to_string());
+/// assert_eq!(distance, 2);
+/// ```
+    pub fn distance_with_adjacent_transposition(a: String, b: String) -> usize {
```
TheElectronWill: Nice doc :)

TheElectronWill: However: why does the function take `String` and not `&str`?

TheElectronWill: once the function accepts `&str`, remove the `to_string()` from the doc

app-agent/src/exec_process.rs
```diff
@@ -12,15 +17,33 @@ use alumet::{
...
+                log::error!("{}", return_error);
+                return Err(e.into());
+            }
+            std::io::ErrorKind::PermissionDenied => {
+                let return_error: String = handle_permission_denied(external_command);
+                log::error!("{}", return_error);
+                return Err(e.into());
+            }
+            _ => {
+                return Err(anyhow!("Error in child process"));
```
TheElectronWill, line 43:
```suggestion
                return Err(anyhow!("Error in child process: {e:?}"));
```

app-agent/src/exec_process.rs
TheElectronWill: @AngeCyp

TheElectronWill (changes requested):
Please propagate errors everywhere. Also, see my comment about the permissions function (numbers vs strings).

app-agent/Cargo.toml
```diff
@@ -26,6 +26,7 @@ plugin-relay = { path = "../plugin-relay", optional = true }
 plugin-procfs = { path = "../plugin-procfs", optional = true }
 plugin-cgroupv2 = { path = "../plugin-cgroupv2", optional = true }
 plugin-influxdb = { path = "../plugin-influxdb", optional = true }
+tempfile = "3.15.0"
```
TheElectronWill, line 29: Put test-only dependencies in `dev-dependencies`:
```toml
[dev-dependencies]
tempfile = "3.15"
```

app-agent/src/exec_process.rs
```diff
@@ -12,15 +17,33 @@ use alumet::{
...
+        .args(args.clone())
+        .spawn();
+
+    let mut p = match p_result {
+        Ok(val) => val,
+        Err(e) => match e.kind() {
+            std::io::ErrorKind::NotFound => {
+                let return_error: String = handle_not_found(external_command, args);
+                log::error!("{}", return_error);
+                return Err(e.into());
```
TheElectronWill, line 35: The error should be logged when it is handled, not here IMO. (=> remove log::error)

app-agent/src/exec_process.rs
```diff
@@ -12,15 +17,33 @@ use alumet::{
...
+            }
+            std::io::ErrorKind::PermissionDenied => {
+                let return_error: String = handle_permission_denied(external_command);
+                log::error!("{}", return_error);
```
TheElectronWill, line 39: remove log error

app-agent/src/exec_process.rs
```diff
@@ -33,6 +56,165 @@ pub fn exec_child(external_command: String, args: Vec<String>) -> anyhow::Result
...
+
+fn find_a_parent_with_perm_issue(path: String) -> anyhow::Result<std::path::PathBuf, String> {
+    // Current parent can change if a parent of the parent don't have the correct rights
+    let mut current_parent = match std::path::Path::new(&path).parent() {
+        Some(parent) => parent,
+        None => return Err("".to_string()),
+    };
+    // Through this loop I will iterate over parent of parent until I can retrieve metadata, it will show the first folder
+    // that I can't execute and suggest to the user to grant execution rights.
+    loop {
```
TheElectronWill, line 205: Isn't this a bit dangerous? We can make it explode with very long paths. Maybe we should just give up and let the user figure it out? Or at least put a loop limit?

TheElectronWill:
Please squash the commits and use `agent` as a scope in the commit messages.

TheElectronWill (changes requested):
Almost done :) I have one remark regarding the git history.

TheElectronWill (approved):

PR #73, opened 2024-12-10
feat(options): config can be passed as environment variable now

### LINKED ISSUE #59: Allow the config path to be defined by an environment variable
```
export ALUMET_CONFIG_PATH=/etc/alumet/config.toml
alumet-local-agent
```

This would be handy for packaged installation of Alumet.

AngeCyp:
Fix #59
You can now use both:

- *--config path/to/config/file*
- *export CONFIG=path/to/config/file*

To use alumet

### DIFF
--- a/app-agent/src/options.rs
+++ b/app-agent/src/options.rs
@@ -56,7 +56,7 @@ pub mod cli {
     #[derive(Args, Clone)]
     pub struct CommonArgs {
         /// Path to the config file.
-        #[arg(long, default_value = "alumet-config.toml")]
+        #[arg(long, env = "ALUMET_CONFIG", default_value = "alumet-config.toml")]
         pub config: String, // not used in Configurator, but directly by main()
 
         /// If set, the config file must exist, otherwise the agent will fail to start with an error.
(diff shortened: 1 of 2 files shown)


app-agent/src/options.rs
```diff
@@ -56,7 +56,7 @@ pub mod cli {
     #[derive(Args, Clone)]
     pub struct CommonArgs {
         /// Path to the config file.
-        #[arg(long, default_value = "alumet-config.toml")]
+        #[arg(long, env, default_value = "alumet-config.toml")]
```
titouanj, line 59: I think it's risky to use CONFIG to store the config path for alumet

titouanj (changes requested):

app-agent/src/options.rs
titouanj: Alumet won't be the only software running on the device i.e. probably not the only one to use CONFIG

titouanj: You should use ALUMET_CONFIG or something like that

TheElectronWill: yes, good point

TheElectronWill: The following will probably work: `env("ALUMET_CONFIG")`

PR #76, opened 2025-01-06
Feature/generic config plugin attribution

cuttingl:


plugin-energy-attribution/src/lib.rs
```diff
@@ -93,13 +99,15 @@ impl AlumetPlugin for EnergyAttributionPlugin {
 
 #[derive(Deserialize, Serialize)]
 struct Config {
-    oui: String,
+    consumed_energy: String,
+    hardware_usage: String,
```
titouanj, line 103: Maybe we should add _metric_name suffix?

titouanj: Or at least change the config key in the config file to make it clearer for the users

plugin-energy-attribution/src/lib.rs
```diff
@@ -26,8 +29,8 @@ struct Metrics {
     //
     // Their IDs are gathered in different phases of the plugin initialization,
     // that is why they are Options.
```
titouanj, line 31: You should update the comment above to make it more generic

plugin-energy-attribution/src/lib.rs
```diff
@@ -44,14 +47,17 @@ impl AlumetPlugin for EnergyAttributionPlugin {
...
+        let config = deserialize_config(config)?;
         Ok(Box::new(EnergyAttributionPlugin {
+            config: Some(config),
             metrics: Arc::new(Mutex::new(Metrics::default())),
         }))
     }
 
     fn start(&mut self, alumet: &mut alumet::plugin::AlumetPluginStart) -> anyhow::Result<()> {
         let mut metrics = self.metrics.lock().unwrap();
+        let config: Config = self.config.take().unwrap();
```
titouanj, line 60: Why do you need to use take?

plugin-energy-attribution/src/transform.rs
```diff
@@ -13,16 +13,19 @@ use alumet::{
...
 
+use crate::Config;
+
 pub struct EnergyAttributionTransform {
+    pub config: Config,
     pub metrics: Arc<Mutex<super::Metrics>>,
     buffer_pod: HashMap<u64, Vec<MeasurementPoint>>,
     buffer_rapl: HashMap<u64, MeasurementPoint>,
 }
-
```
titouanj, line 21: ?

plugin-energy-attribution/src/transform.rs
```diff
@@ -13,16 +13,19 @@ use alumet::{
     resources::Resource,
 };
 
+use crate::Config;
+
 pub struct EnergyAttributionTransform {
+    pub config: Config,
```
titouanj, line 19: You don't need this field here

titouanj (changes requested):

plugin-energy-attribution/src/lib.rs
cuttingl: We need a Config type in the "transform" function (cf transform.rs)

self.config is of type "Option<Config>", we then need to unwrap this value in order to get the "Config" value in itself.
In that case, we can either cast the "self.config" with "self.config.as_ref()" or use the "take()"

plugin-energy-attribution/src/lib.rs
cuttingl: > Maybe we should add _metric_name suffix?

isn't it going against our generic programming process ? or you meant adding the suffix using a dedicated function ?

TheElectronWill: I think Titouan was thinking about the names of the variables: `consumed_energy_metric_name`

plugin-energy-attribution/src/lib.rs
```diff
@@ -67,22 +72,27 @@ impl AlumetPlugin for EnergyAttributionPlugin {
...
-
+        // Finds the RawMetricId with the name of the metric.
+        // Will only run once, just before the pipeline starts.
         // Lock the metrics mutex to apply its modifications.
         let mut metrics = self.metrics.lock().unwrap();
-
-        metrics.rapl_consumed_energy = Some(find_metric_by_name(alumet, "rapl_consumed_energy")?);
-        metrics.cpu_usage_per_pod = Some(find_metric_by_name(alumet, "cgroup_cpu_usage_user")?);
+        let tmp_metrics_consumed_energy = alumet.metrics();
+        let tmp_metrics_hardware_usage = alumet.metrics();
```
TheElectronWill, line 80: You don't need to call `metrics()` multiple times.

plugin-energy-attribution/src/lib.rs
```diff
@@ -67,22 +72,27 @@ impl AlumetPlugin for EnergyAttributionPlugin {
...
+
+        metrics.consumed_energy = Some(
+            tmp_metrics_consumed_energy
+                .by_name(&self.config.consumed_energy)
+                .take()
+                .unwrap()
+                .0,
```
TheElectronWill, line 87: That's a bit convoluted IMO. Also, you should return an error message if the metric is not found (use `context` or `with_context`)

plugin-energy-attribution/src/lib.rs
```diff
@@ -67,22 +72,27 @@ impl AlumetPlugin for EnergyAttributionPlugin {
...
-                .into_iter()
-                .find(|m| m.1.name == name)
-                .unwrap_or_else(|| panic!("Cannot find metric {name}, are the 'rapl' and 'k8s' plugins loaded?"));
-            Ok(id.to_owned())
-        }
-
+        // Finds the RawMetricId with the name of the metric.
+        // Will only run once, just before the pipeline starts.
         // Lock the metrics mutex to apply its modifications.
         let mut metrics = self.metrics.lock().unwrap();
```
TheElectronWill, line 78: You didn't write this part but it can be improved: the Alumet framework now allows transform builders to access the MetricRegistry. Use `add_transform_builder` and get your metrics there, no need for a lock. (when this code was written it was not possible)

TheElectronWill (changes requested):

plugin-energy-attribution/src/lib.rs
TheElectronWill: But: see my comment about the "builder" below: `alumet.metrics()` will be gone

plugin-energy-attribution/src/lib.rs
cuttingl:
```rust
consumed_energy_rapl: String,
hardware_usage_cgroup: String,
``` 

lines changed

plugin-energy-attribution/src/lib.rs
cuttingl:
```rust
    // - overall consumed energy per pod
    // - overall hardware usage per pod
    // - energy attributed to a pod
```

TheElectronWill (changes requested):

plugin-energy-attribution/src/lib.rs
cuttingl: as of this #[c263a51](https://github.com/alumet-dev/alumet/pull/76/commits/c263a5169823c988dfa856deda57ab822310f979) commit, builder is now used instead of the "pre pipeline start"

plugin-energy-attribution/src/transform.rs
cuttingl: metrics are no longer "Arc ... metrics", they are "Metrics" instead

plugin-energy-attribution/src/transform.rs
cuttingl: it has been removed

plugin-energy-attribution/src/lib.rs
cuttingl: those lines are irrelevant now

plugin-energy-attribution/src/lib.rs
cuttingl: same here, these lines are irrelevant

plugin-energy-attribution/src/transform.rs
```diff
@@ -13,15 +12,16 @@ use alumet::{
...
-    pub metrics: Arc<Mutex<super::Metrics>>,
+    pub metrics: super::Metrics,
     buffer_pod: HashMap<u64, Vec<MeasurementPoint>>,
     buffer_rapl: HashMap<u64, MeasurementPoint>,
 }
-
 impl EnergyAttributionTransform {
     /// Instantiates a new EnergyAttributionTransform with its private fields initialized.
-    pub fn new(metrics: Arc<Mutex<super::Metrics>>) -> Self {
+    pub fn new(metrics: Metrics) -> Self {
```
TheElectronWill, line 24:
```suggestion
    pub fn new(metrics: super::Metrics) -> Self {
```
Choose between importing `super::Metrics` or using its full path (line 18 uses `super::Metrics`, line 24 uses `Metrics`)

plugin-energy-attribution/src/transform.rs
```diff
@@ -34,8 +34,8 @@ impl EnergyAttributionTransform {
         // Retrieving the metric_id of the energy attribution.
         // Using a nested scope to reduce the lock time.
         let metric_id = {
-            let metrics = self.metrics.lock().unwrap();
-            metrics.pod_attributed_energy.unwrap()
+            let metrics = &self.metrics;
+            metrics.pod_attributed_energy
```
TheElectronWill, line 38: Suggestion:
```rs
let metric_id = self.metrics.pod_attributed_energy;
```

TheElectronWill (changes requested):
otherwise LGTM

plugin-energy-attribution/src/transform.rs
cuttingl: done in the last commit

plugin-energy-attribution/src/transform.rs
cuttingl: done in the last commit and import (metrics structure) deleted

TheElectronWill (approved):

PR #86, opened 2025-01-16
plugin-tdp: Create metric using the transform builder

cuttingl:
transform builder context added to the tdp plugin transform

### DIFF
--- a/plugin-energy-estimation-tdp/src/lib.rs
+++ b/plugin-energy-estimation-tdp/src/lib.rs
@@ -1,14 +1,16 @@
 use alumet::{
     metrics::{RawMetricId, TypedMetricId},
+    pipeline::elements::transform::builder::TransformRegistration,
     plugin::{
         rust::{deserialize_config, serialize_config, AlumetPlugin},
-        AlumetPreStart, ConfigTable,
+        ConfigTable,
     },
     units::Unit,
 };
 use serde::{Deserialize, Serialize};
-use std::sync::{Arc, Mutex};
 use std::time::Duration;
+
+use anyhow::Context;
 
 use transform::EnergyEstimationTdpTransform;
 
@@ -16,18 +18,16 @@ mod transform;
 
 pub struct EnergyEstimationTdpPlugin {
     config: Option<Config>,
-    metrics: Arc<Mutex<Metrics>>,
 }
 
-#[derive(Default)]
 struct Metrics {
     // To attribute the CPU consumption to K8S pods, we need 2 metrics:
     // - cpu usage per pod
     // - energy attribution (to store the result)
 
     // The other parameters (tdp and number of virtual cpu is provided by configuration)
-    cpu_usage_per_pod: Option<RawMetricId>,
-    pod_estimate_attributed_energy: Option<TypedMetricId<f64>>,
+    cpu_usage_per_pod: RawMetricId,
+    pod_estimate_attributed_energy: TypedMetricId<f64>,
 }
 
 impl AlumetPlugin for EnergyEstimationTdpPlugin {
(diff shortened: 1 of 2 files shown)


plugin-energy-estimation-tdp/src/lib.rs
```diff
@@ -48,48 +48,38 @@ impl AlumetPlugin for EnergyEstimationTdpPlugin {
...
+    
         // Create the energy attribution metric and add its id to the
         // transform plugin metrics' list.
-        metrics.pod_estimate_attributed_energy = Some(alumet.create_metric(
+        let pod_estimate_attributed_energy_metric = alumet.create_metric(
             "pod_estimate_attributed_energy",
             Unit::Joule,
-            "Energy consumption estimated to the pod",
-        )?);
+            "Pod's estimated energy consumption",
```
TheElectronWill, line 61: thanks for the correction

plugin-energy-estimation-tdp/src/lib.rs
```diff
@@ -48,48 +48,38 @@ impl AlumetPlugin for EnergyEstimationTdpPlugin {
...
 
-        // Add the transform now but fill its metrics later.
-        alumet.add_transform(Box::new(EnergyEstimationTdpTransform::new(
-            config,
-            self.metrics.clone(),
-        )));
-        Ok(())
-    }
+        let cpu_usage = self.config.cpu_usage_per_pod.clone();
+        let config_cpy = self.config.clone();
```
TheElectronWill, line 65: Since we don't need the config afterwards, it's time to use `Option::take`: put the config in an option, and `take` it here instead of cloning.

plugin-energy-estimation-tdp/src/lib.rs
cuttingl: 'welcome !

plugin-energy-estimation-tdp/src/lib.rs
cuttingl: we actually use it (transform.rs : line 60 to 63)

```rust
let mut estimated_energy = value.parse().unwrap();
                estimated_energy = estimated_energy * self.config.nb_vcpu / self.config.nb_cpu * self.config.tdp
                    / (1000000.0)
                    / (self.config.poll_interval.as_secs() as f64);
```

cuttingl: done

plugin-energy-estimation-tdp/src/lib.rs
```diff
@@ -48,48 +48,37 @@ impl AlumetPlugin for EnergyEstimationTdpPlugin {
...
 
-        // Add the transform now but fill its metrics later.
-        alumet.add_transform(Box::new(EnergyEstimationTdpTransform::new(
-            config,
-            self.metrics.clone(),
-        )));
-        Ok(())
-    }
+        let cpu_usage = self.config.as_ref().unwrap().cpu_usage_per_pod.clone();
+        let config_cpy = self.config.take().unwrap();
```
TheElectronWill, line 64:
```suggestion
        let config = self.config.take().unwrap();
```
It's not a copy anymore :)

plugin-energy-estimation-tdp/src/lib.rs
```diff
@@ -48,48 +48,37 @@ impl AlumetPlugin for EnergyEstimationTdpPlugin {
...
-        let mut metrics = self.metrics.lock().unwrap();
-
-        metrics.cpu_usage_per_pod = Some(find_metric_by_name(alumet, "cgroup_cpu_usage_total")?);
+        // Add the transform now but fill its metrics later.
+        alumet.add_transform_builder(move |ctx| {
+            let name = ctx.transform_name("energy_estimation_tdp");
+
+            let cpu_usage_metric = ctx
+                .metric_by_name(&cpu_usage)
+                .with_context(|| format!("Metric not found : {}", cpu_usage))?
```
TheElectronWill, line 72: nitpick: error contexts are usually lowercase (by convention)
```suggestion
                .with_context(|| format!("metric not found : {}", cpu_usage))?
```

plugin-energy-estimation-tdp/src/transform.rs
```diff
@@ -34,13 +31,13 @@ impl Transform for EnergyEstimationTdpTransform {
         log::trace!("enter in apply transform function");
 
         let pod_id = {
-            let metrics = self.metrics.lock().unwrap();
-            metrics.cpu_usage_per_pod.unwrap().as_u64()
+            let metrics = &self.metrics;
+            metrics.cpu_usage_per_pod.as_u64()
         };
```
TheElectronWill, line 36: suggestion:
```rs
let metric_cpu_usage = self.metrics.cpu_usage_per_pod.as_u64();
```
idem for `metric_id`, which should be renamed to `metric_attributed_energy` (or something like that)

TheElectronWill (changes requested):
almost done!

TheElectronWill:
Remark about the commit history: you should use `plugin-energy-estimation-tdp` as a scope. So the commit message would start by `feat(plugin-energy-estimation-tdp)`

plugin-energy-estimation-tdp/src/lib.rs
cuttingl: done

plugin-energy-estimation-tdp/src/lib.rs
cuttingl: done

plugin-energy-estimation-tdp/src/transform.rs
cuttingl: old code block, was already the case

TheElectronWill:
Please rebase on the master branch of upstream. If the upstream remote (alumet-dev/alumet) is named `upstream` in your git config, do:
```
git fetch --all
git rebase upstream/main
```

plugin-energy-estimation-tdp/src/transform.rs
```diff
@@ -34,13 +31,13 @@ impl Transform for EnergyEstimationTdpTransform {
...
 
         let metric_id = {
-            let metrics = self.metrics.lock().unwrap();
-            metrics.pod_estimate_attributed_energy.unwrap()
+            let metrics = &self.metrics;
+            metrics.pod_estimate_attributed_energy
         };
```
TheElectronWill, line 41: suggestion: do it in one line
```
let metric_pod_energy = self.metrics.pod_estimate_attributed_energy;
```
similar for `pod_id`

cuttingl: done (with no git issue whatsoever)

TheElectronWill (approved):

PR #93, opened 2025-01-30
The one and only standard agent - unify and expand 🚀

### LINKED ISSUE #54: Provide a way to list the available plugins
It would be nice to have a CLI command that prints the list of available plugins.

### LINKED ISSUE #82: Combine the agents into one
As discussed in meetings, it would be easier to handle a single standard agent (with options of course) than 3 separate ones that are similar-but-not-actually-the-same.

### LINKED ISSUE #91: Provide a clean API for creating agents
Related to #82

TheElectronWill:
Unify: merge all 3 agents into a single one
Expand: new API (in alumet core) for building agents, new capabilities in the new agent

Resolve #54, resolve #82, resolve #91

This will make packaging and tutorials easier.
I mark this on draft so that it stays open at least until next week, but you can still review now.

## Reviewer hint

Read `agent/src/bin/main.rs`: it contains the new Alumet agent.
Everything else is either:
- necessary modifications demanded by the new agent
- unit test / integration test updates
- minor style fixes


TheElectronWill:
I've disabled the musl build test because it doesn't work with `plugin-influxdb`:
```
warning: openssl-sys@0.9.104: Could not find directory of OpenSSL installation, and this `-sys` crate cannot proceed without this knowledge. If OpenSSL is installed and this crate had trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the compilation process. See stderr section below for further information.
error: failed to run custom build command for `openssl-sys v0.9.104`
...


  Could not find openssl via pkg-config:
  pkg-config has not been configured to support cross-compilation.

  Install a sysroot for the target platform and configure it via
  PKG_CONFIG_SYSROOT_DIR and PKG_CONFIG_PATH, or install a
  cross-compiling wrapper for pkg-config and set it via
  PKG_CONFIG environment variable.

  cargo:warning=Could not find directory of OpenSSL installation, and this `-sys` crate cannot proceed without this knowledge. If OpenSSL is installed and this crate had trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the compilation process. See stderr section below for further information.

  --- stderr


  Could not find directory of OpenSSL installation, and this `-sys` crate cannot
  proceed without this knowledge. If OpenSSL is installed and this crate had
  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
  compilation process.

  Make sure you also have the development packages of openssl installed.
  For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.

  If you're in a situation where you think the directory *should* be found
  automatically, please open a bug at https://github.com/sfackler/rust-openssl
  and include information about your system as well as this message.

  $HOST = x86_64-unknown-linux-gnu
  $TARGET = x86_64-unknown-linux-musl
  openssl-sys = 0.9.104
```

What is weird is that changing the `reqwest` dependency to this doesn't not resolve the issue, it still tries to compile openssl bindings.
```toml
reqwest = { version = "0.12.12", default-features = false, features = ["rustls-tls"] }
```

AngeCyp (changes requested):

AngeCyp (approved):

PR #99, opened 2025-02-03
make measurement add_attr public 

guigomcha:
make add_attr public to be able to add new attributes to existing measurement points during a transform.


TheElectronWill:
Thanks for the PR!
Please keep only the modifications to `alumet/src/measurements.rs`. (the agent will change with #93 anyway)

alumet/src/measurement.rs
```diff
@@ -130,14 +130,16 @@ impl MeasurementPoint {
...
-    pub(crate) fn add_attr(&mut self, key: Cow<'static, str>, value: AttributeValue) {
-        self.attributes.push((key, value));
+    /// Sets an attribute on this measurement point.
+    /// If an attribute with the same key already exists, its value is replaced.
+    pub fn add_attr<K: Into<Cow<'static, str>>, V: Into<AttributeValue>>(&mut self, key: K, value: V) {
+        self.attributes.push((key.into(), value.into()));
     }
 
-    /// Sets an attribute on this measurement point.
+    /// Sets an attribute and returns a new measurement point.
```
TheElectronWill, line 139: Actually it does not return a new measurement point but this one. I would describe it as:
```rs
/// Sets an attribute on this measurement point, and returns self to allow for method chaining.
```

TheElectronWill (changes requested):

guigomcha:
> Thanks for the PR! Please keep only the modifications to `alumet/src/measurements.rs`. (the agent will change with #93 anyway)

oka, I did not see that one. Please consider the changes anyway if they apply in the new version

alumet/src/measurement.rs
guigomcha: I meant to express that the original self is unmodified and that is why we can only use this method during creation via method chaining. Updated with your contribution anyway :)

guigomcha:
@TheElectronWill I usually wait until the CI is ok to set the PR as ready and add reviewers but other people make it more explicit on the PR name. How do you guys work?

guigomcha:
btw, @TheElectronWill the CI remains blocked until I manually approve the job. Is that the expected behavior?
image

.gitignore
```diff
@@ -6,4 +6,4 @@ app-*/perf.data
 app-*/perf.data.old
 app-*/*output.csv
 .DS_Store
-.vscode
+.vscode
```
TheElectronWill, line 9: Why remove the trailing line?

TheElectronWill:
> btw, @TheElectronWill the CI remains blocked until I manually approve the job. Is that the expected behavior? I do not have an account and cannot approve. Is that an enterprise account or public?

Some jobs require an approval from the maintainers (to avoid CI spamming), I have approved it :)

.gitignore
guigomcha, line 9:
```suggestion
.vscode

```

PR #116, opened 2025-02-18
Prometheus exporter plugin

### LINKED ISSUE #7: Implement a Prometheus output
Questions to answer in meeting:
- [ ] push or pull?
- [ ] which library to use?
- [ ] which config options to offer?

guigomcha:
Plugin to expose the metrics following the OpenMetrics specification which can be connected to prometheus and to opentelemetry

Resolves #7


guigomcha:
@TheElectronWill  I think it is ready for review

.cspell/project-words.txt
```diff
@@ -27,6 +27,7 @@ NVML
 pagetables
 PERFMON
 POWERCAP
+promstack
```
TheElectronWill, line 30: I don't like the idea of adding non-words to the dictionary. Could you ignore the file instead?

.gitignore
```diff
@@ -6,3 +6,4 @@ app-*/perf.data.old
 app-*/*output.csv
 .DS_Store
 .vscode
+**/debug/**
```
TheElectronWill, line 9: This can bite us later if we add a `debug` Rust module as a folder. What are you trying to ignore?
You can use `target` or `tmp` for testing.

plugin-prometheus-exporter/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+use output::PrometheusOutput;
+use serde::{Deserialize, Serialize};
+
+pub struct PrometheusPlugin {
+    config: Config,
+}
+
+impl AlumetPlugin for PrometheusPlugin {
+    fn name() -> &'static str {
+        "plugin-prometheus"
```
TheElectronWill, line 13:
```suggestion
        "prometheus-exporter"
```

plugin-prometheus-exporter/src/output.rs
```diff
@@ -0,0 +1,216 @@
...
+        let registry = Arc::new(RwLock::new(Registry::default()));
+        let metrics = Arc::new(RwLock::new(HashMap::new()));
+        let state = MetricState { registry, metrics };
+
+        // Configure the HTTP server to expose the metrics
+        let addr: SocketAddr = format!("{}:{}", host, port)
+            .parse()
+            .context("Invalid host:port configuration")?;
+
+        let rt = Runtime::new().context("Failed to create Tokio runtime")?;
```
TheElectronWill, line 59: I think you can just use the Alumet async runtime instead of creating a new one in a new thread (accessible in the context provided by `alumet.on_pipeline_start` or `alumet.add_async_output_builder`).

You also need to handle the graceful shutdown of the output: when the prometheus output is stopped, the hyper server must stop too. This one is a bit more tedious with the current core: you will have to
- turn your blocking output into an async one
- detect when the output is shut down (the `stream` is closed, i.e. `stream.0.next()` will return `None`)
- send a signal to the hyper service (or put the 

```rs
    alumet.add_async_output_builder("prometheus_exporter", move |ctx, stream| {
        let signal = todo!(); // you can use Notify or CancellationToken
        let alumet_output = async move {
            loop {
                match stream.0.next() {
                    Some(Ok(buf)) => todo!("update prometheus metrics"),
                    Some(Err(e)) => todo!("error"),
                    None => todo!("shutdown the http server") // use the signal here
                }
            }
        };

        let make_svc = make_service_fn(|_conn| todo!());
        let http_server = Server::bind(addr).serve(make_svc);
        let http_server = server.with_graceful_shutdown(signal);
        // spawn the http server on the already existing async runtime
       ctx.async_runtime().spawn(http_server)?;

        // let Alumet manage the async output
        let output: BoxedAsyncOutput = Box::pin(alumet_output);
        Ok(output)
    });
```

.gitignore
guigomcha: I have built the binary to use it elsewhere and was left with over 14k files in my untracked files. Maybe just put t in agent/debug/**?

btw I think that the alumet-output.csv was tracked after the migration to the new agent folder due to the line 6 of this gitignore

plugin-prometheus-exporter/src/output.rs
guigomcha: ok, thanks for the info. This is gonna take me a bit more to digest and apply xD

.cspell/project-words.txt
guigomcha: fixed

plugin-prometheus-exporter/src/output.rs
guigomcha: btw @TheElectronWill , I used the influxdb and the mongodb plugins as reference to implement it in that way. Would this be a nice-to-have or something that we expect to push plugins to use more? The only reference that I see is the relay plugin and is extremely more complicated to follow the logic

If the API is going to be updated to be more friendly in this cases I would prefer to wait and upgrade the plugin afterwards

.gitignore
TheElectronWill: The binary should end up in target/debug or target/release, not in the agent directory 🤔 We use a cargo workspace.

plugin-prometheus-exporter/src/output.rs
TheElectronWill: Well it depends on the plugin. The influxdb and mongodb did not start a background async task, which makes a big difference in how they are registered into Alumet pipeline ^^.

.gitignore
guigomcha: mmm I have been running it inside the agent folder, my bad!

.cspell.json
```diff
@@ -14,6 +14,7 @@
         ".git/*",
         ".gitignore",
         ".github/",
+        "plugin-prometheus-exporter/docs/alumet-scrape-config.yaml",
```
TheElectronWill, line 17: All toml files are ignored, let's ignore all yaml too.

TheElectronWill (commented):
Please squash fixes in the first commit and use the following commit convention (derived from [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/))

```
type(scope): message
```
where `scope` is `plugin-prometheus-exporter` and `type` is one of: feat, fix, doc, refactor

If the type is unclear you can keep the scope only:
```
scope: message
```

plugin-prometheus-exporter/README.md
```diff
@@ -0,0 +1,42 @@
+# Prometheus Exporter plugin
+
+This crate is a library that defines the Prometheus Exporter plugin.
+
+Implements a pull-based exporter which can be consumed by all monitoring tools compatible with the OpenMetrics specification.
+
+## Demos
+
+The plugin has been tested on a NUC with the default configuration.
```
guigomcha, line 9:
```suggestion
The plugin has been tested on a NUC with the default configuration.
```

guigomcha:
> Please squash fixes in the first commit and use the following commit convention (derived from [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/))
> 
> ```
> type(scope): message
> ```
> 
> where `scope` is `plugin-prometheus-exporter` and `type` is one of: feat, fix, doc, refactor
> 
> If the type is unclear you can keep the scope only:
> 
> ```
> scope: message
> ```

The PR has been automatically closed due to the rebase and resync with the main branch. The same PR is here now https://github.com/alumet-dev/alumet/pull/123 

btw, I believe that you can configure PRs to only accept squash commits on merge (which is done automatically) and that way it would simplify a lot the contribution process

guigomcha:
reopening after squash

PR #117, opened 2025-02-20
OpenTelemetry output for OpenSearch demo

### LINKED ISSUE #97: OpenSearch / ElasticSearch plugin
Requested by the LIG for experiments on the Gricad cluster (to start mid-february!).

guigomcha:
Replicated the structure of the prometheus plugin to export directly to opentelemetry.

It has been tested with a local deployment of Opensearch

Resolves #97


guigomcha:
@TheElectronWill I think this is ready for review

TheElectronWill:
Is there a reason for the "97" in the PR title?

guigomcha:
> Is there a reason for the "97" in the PR title?

Hi, yes. I did the dev and demo based on opensearch and I swa this issue https://github.com/alumet-dev/alumet/issues/97  where I was waiting for feedback but in the end I got it to work.

TheElectronWill:
> > Is there a reason for the "97" in the PR title?
> 
> Hi, yes. I did the dev and demo based on opensearch and I swa this issue https://github.com/alumet-dev/alumet/issues/97  where I was waiting for feedback but in the end I got it to work.

Oh, okay! In that case I will edit the title and add a github-specific description to link the PR to the issue.

agent/src/bin/main.rs
```diff
@@ -28,6 +28,7 @@ fn load_plugins_metadata() -> Vec<PluginMetadata> {
         plugin_mongodb::MongoDbPlugin,
         plugin_relay::client::RelayClientPlugin,
         plugin_relay::server::RelayServerPlugin,
+        plugin_opentelemetry::OpentelemetryPlugin,
```
AngeCyp, line 31:
```suggestion
        plugin_opentelemetry::OpentelemetryPlugin,
        plugin_relay::client::RelayClientPlugin,
        plugin_relay::server::RelayServerPlugin,
```

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+impl alumet::pipeline::Output for OpentelemetryOutput {
+    fn write(&mut self, measurements: &MeasurementBuffer, ctx: &OutputContext) -> Result<(), WriteError> {
+        if measurements.is_empty() {
+            return Ok(());
+        }
+        // Needs to be created inside the tokio thread
+        let meter_provider = init_metrics();
```
AngeCyp, line 68: As a question, can't you just build the SdkMeterProvider once and reuse it ?

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,144 @@
...
+        global::set_meter_provider(meter_provider.clone());
+        let common_scope_attributes = vec![KeyValue::new("tool", "alumet")];
+        let scope = InstrumentationScope::builder("alumet")
+            .with_version("1.0")
```
AngeCyp, line 72: Is it always 1.0 ?

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,144 @@
...
+            // Prepare the meter provider
+            let meter = global::meter_with_scope(scope.clone());
+            let gauge = meter
+                .f64_gauge(metric_name)
+                .with_description(metric.description.to_string())
+                .with_unit(metric.unit.display_name())
+                .build();
+            match m.value {
+                WrappedMeasurementValue::F64(v) => gauge.record(v as f64, &labels),
+                WrappedMeasurementValue::U64(v) => gauge.record(v as f64, &labels),
```
AngeCyp, line 131: Is the "as f64" always needed ?

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,144 @@
...
+impl OpentelemetryOutput {
+    pub fn new(
+        append_unit_to_metric_name: bool,
+        use_unit_display_name: bool,
+        add_attributes_to_labels: bool,
+        collectot_host: String,
+        prefix: String,
+        suffix: String,
+    ) -> anyhow::Result<OpentelemetryOutput> {
+        env::set_var(
```
TheElectronWill, line 48: What is it for?
`env::set_var` is now `unsafe` because it is not ok to call it in a multithreaded environment.

plugin-opentelemetry/src/output.rs
TheElectronWill: Yes, I see multiple structs and vars that should be initialized only once, when creating the `OpentelemetryOutput`.

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,144 @@
...
+};
+use anyhow::Context;
+use opentelemetry::{global, InstrumentationScope, KeyValue};
+use opentelemetry_otlp::MetricExporter;
+use opentelemetry_sdk::metrics::SdkMeterProvider;
+use opentelemetry_sdk::Resource;
+use std::{env, sync::OnceLock};
+
+#[derive(Clone)]
+pub struct OpentelemetryOutput {
```
TheElectronWill, line 13:
```suggestion
pub struct OpenTelemetryOutput {
```

TheElectronWill: casing conventions

plugin-opentelemetry/src/output.rs
TheElectronWill: If it needs to be created inside of the tokio thread, you can use `add_blocking_output_builder` and create the provider inside of the builder. If it doesn't work, that's something that I will fix in core :)

plugin-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    fn default_config() -> anyhow::Result<Option<alumet::plugin::ConfigTable>> {
+        Ok(Some(serialize_config(Config::default())?))
+    }
+
+    fn init(config: alumet::plugin::ConfigTable) -> anyhow::Result<Box<Self>> {
+        let config: Config = deserialize_config(config)?;
+        // Create a new OpentelemetryOutput instance
+        let otel_output = Box::new(OpentelemetryOutput::new(
```
TheElectronWill, line 27: It's better to clone/move the config instead of the output (in particular with my remark below: the output should contain more initialized things instead of initializing them in `write`).

TheElectronWill (changes requested):

plugin-opentelemetry/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+anyhow = "1.0.88"
+log = "0.4.22"
+serde = { version = "1.0.210", features = ["derive"] }
+time = { version = "0.3.36", features = ["formatting"] }
+hyper = { version = "0.14", features = ["full"] }
+hyper-tls = "0.5"
+tokio = { version = "1.0", features = ["full"] }
+opentelemetry = { version = "*" }
+opentelemetry_sdk = { version = "*" }
+opentelemetry-otlp = { version = "*", features = ["grpc-tonic"]}
```
TheElectronWill, line 19: The musl build is failing because (at least) one of the dependencies is looking for openssl. Probably hyper, so it will be resolved when you remove it from the dependency list. If opentelemetry does that too, you should tweak its features to use rustls instead openssl on musl target (see `plugin-influxdb` for example).

plugin-opentelemetry/src/output.rs
TheElectronWill: See https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html#stdenvset_var-remove_var

plugin-opentelemetry/src/output.rs
guigomcha: add_blocking_output_builder does not work, I am trying with add_async_output_builder but it is not straigthforward for sure

plugin-opentelemetry/src/output.rs
guigomcha: nop, I changed it for the plugin version

plugin-opentelemetry/src/output.rs
guigomcha: it is needed to create that gauge yes. There are more options but f64 seems flexible enough for now

plugin-opentelemetry/src/output.rs
guigomcha: doe

plugin-opentelemetry/src/output.rs
guigomcha: fixed in new commit

TheElectronWill:
Please rebase instead of merging in order to keep the history clean :)

plugin-opentelemetry/src/lib.rs
TheElectronWill: In the current state of the branch, the output is still stored in the plugin's structure. Has something be lost by the merge?

plugin-opentelemetry/src/output.rs
TheElectronWill: Env vars aren't necessary, it seems that you can use `with_endpoint` on the builder instead:
```rs
    let exporter = opentelemetry_otlp::MetricExporter::builder()
       .with_tonic()
       .with_endpoint("http://localhost:4318/v1/metrics")
       ... // other settings here
      .build()?;
```

TheElectronWill: Note that even if the other threads are reading the env variables, it's _not safe_. Quote from the [docs](https://doc.rust-lang.org/std/env/fn.set_var.html#safety):
> In multi-threaded programs on other operating systems, the only safe option is to not use set_var or remove_var at all.

Alumet makes no guarantee that only one thread is running when the plugins are initialized.

plugin-opentelemetry/src/lib.rs
guigomcha: no

plugin-opentelemetry/src/output.rs
guigomcha: I did not find that method, thanks!.SHould be fixed now

plugin-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+            config.add_attributes_to_labels,
+            config.prefix.clone(),
+            config.suffix.clone(),
+            config.collector_host.clone(),
+        )?);
+        Ok(Box::new(OpenTelemetryPlugin { output: otel_output }))
+    }
+
+    fn start(&mut self, alumet: &mut alumet::plugin::AlumetPluginStart) -> anyhow::Result<()> {
+        alumet.add_blocking_output(OpenTelemetryPlugin::name(), self.output.clone());
```
TheElectronWill, line 39: The String given to `add_blocking_output` is the name of the output, not the plugin. Here you have only one output so it doesn't really matter, I would put something like `"out"`.

plugin-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,70 @@
+mod output;
+
+use alumet::plugin::rust::{deserialize_config, serialize_config, AlumetPlugin};
+use output::OpenTelemetryOutput;
+use serde::{Deserialize, Serialize};
+
+pub struct OpenTelemetryPlugin {
+    output: Box<OpenTelemetryOutput>,
```
TheElectronWill, line 8: Can you put the config here instead of the output, and create the output in `start`?

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,156 @@
...
+            use_unit_display_name,
+            add_attributes_to_labels,
+            prefix,
+            suffix,
+            initialized: false,
+            collector_host: format!("{}{}", collector_host, "/v1/metrics"),
+        })
+    }
+    pub fn initialize(&mut self) {
+        // Needs to be created inside the tokio thread
```
TheElectronWill, line 43: Can you add a TODO so that we remember to change that once #119 is fixed? Ideally, this would be done in the builder, but here it's not possible because output builders lack the runtime context.

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,156 @@
...
+            self.initialize();
+            self.initialized = true;
+        }
+        let common_scope_attributes = vec![KeyValue::new("tool", "alumet")];
+        let scope = InstrumentationScope::builder("alumet")
+            .with_version(env!("CARGO_PKG_VERSION"))
+            .with_attributes(common_scope_attributes)
+            .build();
+        for m in measurements {
+            let metric = ctx.metrics.by_id(&m.metric).unwrap().clone();
```
TheElectronWill, line 83: `metric` seems unused, remove it

plugin-opentelemetry/src/output.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+// Helper functions to ensure metric/label names follow Prometheus naming rules
+fn sanitize_name(name: String) -> String {
+    name.chars()
+        .map(|c| if c.is_ascii_alphanumeric() || c == '_' { c } else { '_' })
```
TheElectronWill, line 154: colons should be allowed too, no?

> Metric names may contain ASCII letters, digits, underscores, and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.

TheElectronWill (changes requested):

plugin-opentelemetry/docs/otel-opensearch-demo.png
TheElectronWill, line 1: After discussing with the team, we think that such screenshots should go to the [User Book](https://github.com/alumet-dev/user-book), not here. You can open a PR on the book's repository.

plugin-opentelemetry/src/lib.rs
guigomcha: the idea is that the name shows that it points to the promethues plugin right? if there were more than one outout I would understand to append "_out" or similar, but just "out" doesn't say much I think

plugin-opentelemetry/src/output.rs
guigomcha: where did you see this?

plugin-opentelemetry/docs/otel-opensearch-demo.png
guigomcha: There is no baseline info on how to cover the explanation of the plugins there and it will take some time. I'll keep the files in the issue I created https://github.com/alumet-dev/user-book/issues/12  in the user book but would prefer not to block this PR more

plugin-opentelemetry/src/lib.rs
guigomcha: done

plugin-opentelemetry/docs/otel-opensearch-demo.png
TheElectronWill: That's good, no need to fill in the user book now.

plugin-opentelemetry/src/output.rs
TheElectronWill: Here: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

plugin-opentelemetry/src/lib.rs
TheElectronWill: `out` is fine because Alumet automatically remembers the plugin name.
For instance, in the logs, it will appear as `outputs/plugin-prometheus/out`

plugin-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+use output::OpenTelemetryOutput;
+use serde::{Deserialize, Serialize};
+
+pub struct OpenTelemetryPlugin {
+    config: Config,
+}
+
+impl AlumetPlugin for OpenTelemetryPlugin {
+    fn name() -> &'static str {
+        "plugin-opentelemetry"
```
TheElectronWill, line 13:
```suggestion
        "opentelemetry"
```
By convention, we don't put "plugin" in the name of the plugin (see other plugins). I should add that to the dev book.

TheElectronWill (changes requested):
almost done

plugin-opentelemetry/src/output.rs
guigomcha: Hi, but they have a special meaning that cannot be used by the exporters 
image

It is true that this is the opentelemetry plugin and they do not specify any limitation, I could just delete it.

plugin-opentelemetry/src/lib.rs
guigomcha: awesome, I did not know that

plugin-opentelemetry/src/output.rs
TheElectronWill: good catch, I was not aware of the special meaning. Do what you think is best then.

plugin-opentelemetry/src/lib.rs
guigomcha: I'll update the prometheus plugin too

TheElectronWill (approved):

Issue #65, opened 2024-11-21
Make Alumet increases its Linux resource limits automatically


TheElectronWill:
**Is your feature request related to a problem? Please describe.**
Context: on some systems, the default `ulimit -Sn` is too low for the `procfs` plugin. See discussion #64.
For now, it's up to the user/admin to configure their system properly.

**Describe the solution you'd like**
The agent could try to increase its soft limit by itself, using [these C functions](https://linux.die.net/man/2/setrlimit).

Alternative: provide packages with systemd services that are configured with an appropriate `LimitNOFILE`.

TheElectronWill:
bullet points everywhere make it seem (to me) like it's AI generated 😄 👀 

Approaches 1 and 2 feel too cumbersome to implement, and approach 3 is not user-friendly enough IMO.
Basic users don't want to learn how fd limits work, why they don't always get the error, etc. Furthermore, the limit system is designed to allow an automatic increase when it is needed.

I would like to advocate for _solution 4_:
We know that the _procfs_ plugin is very likely to hit the limit, because it basically creates one source per process.
The code that create the sources is quite straightforward (a loop on the /proc virtual directory), and it would not be hard to detect when we are going to blow the limit away.
**Even simpler:** just increase the limit automatically, unconditionally, at the beginning of the procfs plugin. The [`rlimit` crate](https://docs.rs/rlimit/latest/rlimit/) makes it easy to implement. AFAIK there is no downside to increasing the soft limit.

ahuret:
Indeed, it looks much more straightforward ! :D

PR #118, opened 2025-02-24
plugin-procfs automatically increases file descriptors soft limit

ahuret:
Automatically increase alumet process's file descriptors soft limit when plugin-procfs starts in order to prevent 'Too many open files' error that happened because procfs plugin opens lot of files.

It starts without condition (we could, later, add a configuration option - plugin side - to disable this behavior in some scenarios where administrator would like to manage this limit precisely).

The current implementation will check for the process hard limit and will update soft limit to reach the hard limit. This way the user has still control on the overall process limit - by defining the hard limit.

A future implementation could also be located in Alumet's core to perform the same operation for all the plugins.

### DIFF
--- a/plugin-procfs/src/lib.rs
+++ b/plugin-procfs/src/lib.rs
@@ -9,6 +9,7 @@ use alumet::{
 };
 use anyhow::Context;
 use procfs::{Current, CurrentSI};
+use rlimit::{getrlimit, setrlimit, Resource};
 
 mod kernel;
 mod memory;
@@ -68,6 +69,8 @@ impl AlumetPlugin for ProcfsPlugin {
                 }
             }
         }
+        increase_file_descriptors_soft_limit().context("Error while increasing file descriptors soft limit")?;
+
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


ahuret:
Closes #65

plugin-procfs/src/lib.rs
```diff
@@ -182,6 +185,15 @@ fn setup_process_event_listener(
     });
 }
 
+// prevent 'Too many open files' error
+fn increase_file_descriptors_soft_limit() -> Result<(), anyhow::Error> {
+    let (fd_soft, fd_hard) = getrlimit(Resource::NOFILE).context("Error while getting file descriptors limits")?;
+    setrlimit(Resource::NOFILE, fd_hard, fd_hard)
```
AngeCyp, line 191: Why two times `fd_hard` ? Couldn't you just use old values and multiply them by X or use hard coded limit ?

TheElectronWill:
The commit message should be:
```
feat(plugin-procfs): automatically increases file descriptors soft limit
```

TheElectronWill (changes requested):
The commit message should be:
```
feat(plugin-procfs): automatically increases file descriptors soft limit
```

TheElectronWill:
Can you rebase on main? If I do it via GitHub, your commit will not be signed.

ahuret:
> Can you rebase on main? If I do it via GitHub, your commit will not be signed.

done !

AngeCyp (approved):

TheElectronWill (approved):

PR #121, opened 2025-03-03
Core interface for end-to-end plugin testing

### LINKED ISSUE #92: API for testing plugins
Testing plugins is quite complicated because it require to setup a full agent and/or to modify the plugin implementation to assert that things are correct. It would be better if Alumet core could provide a unified testing module that would be used by all plugin tests.

Example of a test (brainstorming with @AngeCyp):
```rust

const TIMEOUT: Duration = Duration::from_secs(2);

#[test]
fn plugin_in_pipeline() {
    struct TestedPlugin;

    let tester = alumet::test::RuntimeExpectations::new()
        .source_output("tested/source/1", |m| {
            assert_eq!(m.len(), 2);
            assert_eq!(m[0].value, 123.5);
        })
        .transform_output("t1", |m| {todo!()})
        .transform_result("t1", |input| {input.push(MeasurementPoint::new(...))}, |output| {assert_eq!(output, ...)})
        .build();
    
    let mut plugins = static_plugins![TestedPlugin];
    plugins.push(tester);
    
    let mut plugins = agent::plugin::PluginSet::new(plugins);
    
    let expectations = alumet::test::StartupExpectations::new()
        .start_metric("energy", Metric { ... })
        .start_metric("voltage", Metric { ... })
        .start_metric("cyprien-is-too-fast", Metric { ... })
        .element_source("source1", SourceType::Managed)
        .element_transform("tron")
        .build();

    let agent = agent::Builder::new(plugins)
        .with_expectations(expectations)
        .build_and_start()
        .expect("startup failure");
    
    agent.wait_for_shutdown(TIMEOUT).unwrap();
}
```

TheElectronWill:
The new module `alumet::test` allows to write integration tests for plugins.
The goal is to test plugins in a "realistic" setup with the actual Alumet pipeline, from startup to shutdown.

With the new module, you can declare tests that check input and output of each source, transform and output individually.

Remaining things to do:
- [x] write tests
- [x] make metric testing less cumbersome with a custom structure
- [x] write extensive doc for StartupExpectations
- [x] write extensive doc for RuntimeExpectations
- [x] write extensive doc for the `test` module itself

Note : some hacks are still there, in particular `sleep` calls that wait for control commands to be processed.
To get rid of them, the control API would need to be improved.

Resolves #92


alumet/src/pipeline/elements/error.rs
```diff
@@ -1,3 +1,4 @@
 pub use crate::pipeline::elements::output::error::WriteError;
 pub use crate::pipeline::elements::source::error::PollError;
 pub use crate::pipeline::elements::transform::error::TransformError;
+// TODO remove the above aliases
```
AngeCyp, line 4: Planned for another PR?

AngeCyp (approved):

alumet/src/pipeline/elements/error.rs
TheElectronWill: yes

PR #127, opened 2025-03-19
New control API

TheElectronWill:
# Changes summary

- The structure of the control messages is now hidden behind an API based on the builder pattern.
- It is now possible to wait for a control command to be processed and to get its result.
- Sending a message is now async, because it needs to wait for the buffer to have some space. This need was already there before this PR, but it was ignored (the code panicked when the buffer was full).
- Pause/Resume/Enable/Disable have been harmonized: now, there is simply `enable` and `disable`
- Resolve dirty workaround that was part of new test module

For more details, see rustdoc :)


alumet/src/pipeline/control/main_loop.rs
```diff
@@ -0,0 +1,206 @@
...
+                            if let Err(e) = self.handle_message(msg).await {
+                                log::error!("error in message handling: {e:?}");
+                                last_error = Err(e);
+                            }
+                        },
+                        None => todo!("pipeline_control_loop#rx channel closed"),
+                    }
+                },
+
+                // Below we asynchronously poll the source, transform and output tasks, in order to detect
```
TheElectronWill, line 155: I was bitten by this, so I've added a detailed comment :smile:

plugin-procfs/src/process.rs
```diff
@@ -363,33 +384,35 @@ impl ProcessWatcher {
...
-            .flush()
-            .context("failed to create sources for new processes detected by ProcessWatcher")?;
+        // Send the request and handle errors.
+        let request = request_builder.build();
+
+        // Sources run in a tokio context, we can obtain a Handle to the runtime.
+        let handle = &self.alumet_handle;
+        tokio::runtime::Handle::try_current()
+            .expect("refresh must be called in a tokio runtime")
+            .block_on(handle.dispatch(request, None))?;
```
titouanj, line 401: Seems to introduce a bug 😄

PR #128, opened 2025-03-19
Only enable the plugins that are mentioned in the agent config

### LINKED ISSUE #69: Harmonize "enable/enabled" in config files
It can be unsettling that some options are named `enable` (enables a plugin) and others `enabled` (in procfs plugin).

Choose a consistent naming and use it everywhere. Or, accept both?

### LINKED ISSUE #110: Feature idea : allow plugin selection in config files (very small QoL)
Currently, CLI options allow the user to select the plugins and the config file they wish to use. The thing is, it seems to me that in most cases a config file will always be used with the same plugins, so it would make sense to be able to specify a plugin list in the config file once and for all.

TheElectronWill:
Resolve #110, resolve #69.

Based on #127 (so please merge it first).

### DIFF
--- a/alumet/src/agent/config.rs
+++ b/alumet/src/agent/config.rs
@@ -389,7 +389,7 @@ pub fn merge_override(original: &mut toml::Table, overrider: toml::Table) {
 ///     global_option = "this value is not for plugins"
 ///
 ///     [plugins.a]
-///     enabled = false  # default is enabled (see plugins.b)
+///     enabled = false
 ///     key = "value"
 ///     
 ///     [plugins.b]
@@ -430,7 +430,7 @@ pub fn extract_plugins_config(config: &mut toml::Table) -> Result<BTreeMap<Strin
     ) -> Result<(bool, toml::Table), BadTypeError> {
         match config_section {
             toml::Value::Table(mut plugin_config) => {
-                // get the TOML value
+                // get the TOML value, true if absent
                 let enabled_val = plugin_config
                     .remove("enabled")
                     .or_else(|| plugin_config.remove("enable"))
--- a/alumet/src/agent/plugin.rs
+++ b/alumet/src/agent/plugin.rs
@@ -79,12 +79,8 @@ pub enum PluginFilter {
 pub enum UnknownPluginInConfigPolicy {
     /// Logs a warning message and continues.
     LogWarn,
-    /// Logs a warning message only if the config enables the plugin.
-    LogWarnIfEnabled,
     /// Returns an error.
     Error,
-    /// Returns an error only if the config enables the plugin.
-    ErrorIfEnabled,
     /// Ignores the plugin config and continues.
     Ignore,
 }
@@ -116,6 +112,8 @@ impl PluginSet {
     }
 
     /// Enables the specified plugins and disables all the others.
+    ///
+    /// Plugins that are not in the set are ignored.
     pub fn enable_only(&mut self, plugin_names: &[impl AsRef<str>]) {
         // We disable every plugin and re-enable only the ones we are interested in.
         // Cost: O(P+E) where P is the number of plugins in the set and E the size of `plugin_names`,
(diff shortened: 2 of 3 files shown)


guigomcha:
does this mean that we cannot use the enable=false to disable a plugin when the binary has been created with it?

titouanj (approved):

TheElectronWill:
Actually, I'm thinking about a hybrid approach where we keep the possibility of using a special `enable` key. For instance, it could be useful for creating helm charts (where generating a list of plugins could be painful) or if one wants to use environment variables to enable individual plugins.
```toml
[plugins.xyz]
enable = ${DEPENDS_ON_ENV}
```

What do you think @titouanj?

titouanj:
> Actually, I'm thinking about a hybrid approach where we keep the possibility of using a special `enable` key. For instance, it could be useful for creating helm charts (where generating a list of plugins could be painful) or if one wants to use environment variables to enable individual plugins.
> ```toml
> [plugins.xyz]
> enable = ${DEPENDS_ON_ENV}
> ```
> 
> What do you think @titouanj?

Well, I think it's a good idea to keep this alternative. But I don't get why it could be painful to generate a list when using Helm.

It could be useful to overide the config without the --plugins flag. So let's keep it!

TheElectronWill:
> But I don't get why it could be painful to generate a list when using Helm.

IIRC Manu uses one variable per plugin, so he would have to take every variable and build a string by concatenating small bits before passing them to `--plugins`.

TheElectronWill:
I've modified the PR to allow both methods:
- Using `enable` is still supported, just like before.
- What's new: removing a plugin from the config (including by commenting its config section) disables it;

guigomcha:
> I've modified the PR to allow both methods:
> 
> * Using `enable` / `enabled` is still supported, just like before.
> * What's new: removing a plugin from the config (including by commenting its config section) disables it;

In fact, would it be possible to automatically inject the enable = true to the config sections? I tried to add an enable in my plugins what it gave me some trouble

TheElectronWill:
Why do you want to inject `enable = true` in the config?

titouanj:
> > But I don't get why it could be painful to generate a list when using Helm.
> 
> IIRC Manu uses one variable per plugin, so he would have to take every variable and build a string by concatenating small bits before passing them to `--plugins`.

Maybe. But that's not that complicated to do 😄
(it may be interesting to rework the helm charts if such small things is hard to change)

guigomcha:
> Why do you want to inject `enable = true` in the config? If it's not present, Alumet sets the flag to `true` internally.

I have found myself manually adding the enable = false many times because I didn't want to comment out lines in the agent main.rs but I cannot execute many plugins that are enabled by default (rapl, influxdb etc....). 
Also, the config file is not recreated on each build if it already exists, which means that when you coment a plugin in the agent's main.rs, you also need to remove its config or add the enabled=false to make sure that you don't get an error

TheElectronWill:
The way that it will work with this PR is:
- every plugin that appears in the config is enabled...
- ...except if its config section contains `enabled = false`
- the list of enabled plugins is overriden by the CLI argument `--plugins`

So, with this PR, you will simply have to remove/comment the plugins that you don't want from the config, and they will be disabled. Note that you can generate a config that contains only some plugins with `alumet-agent config regen --plugins a,b,c`

guigomcha:
> The way that it will work with this PR is:
> 
> * every plugin that appears in the config is enabled...
> * ...except if its config section contains `enabled = false`
> * the list of enabled plugins is overriden by the CLI argument `--plugins`
> 
> So, with this PR, you will simply have to remove/comment the plugins that you don't want from the config, and they will be disabled. Note that you can generate a config that contains only some plugins with `alumet-agent config regen --plugins a,b,c`.

oook, I didn't use the --plugin flag in the regen comand, it might be good to add it here https://alumet-dev.github.io/user-book/installation/config.html#regenerating-the-file

Issue #115, opened 2025-02-18
Provide package with every release


AngeCyp:
For each GitHub release, it would be great to attach:

-  The binaries of the Alumet agents
-  RPM packages?
-  DEB packages?
- Container image

If it's automatic, it's even better. It would work like this:

1. A new release is created on alumet-dev/alumet
2. A github action runs, builds the RPM and DEB packages
3. Add them to the release

PR #133, opened 2025-03-20
Build/create release action

AngeCyp:
This PR add code for Github action to create all necessary artifacts and attach them to a release. 

- Correctly call reusable workflow from alumet-dev/packaging repo
- Take in count increase of release version on a new build

Close: #115


.github/workflows/release.yaml
titouanj, line 1: Why the tabs?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,142 @@
...
+              RAW_VERSION=${{ github.event.release.tag_name }}
+              CLEAN_VERSION=${RAW_VERSION#v}
+              echo "VERSION=$CLEAN_VERSION" >> $GITHUB_OUTPUT
+              echo "RELEASE=1" >> $GITHUB_OUTPUT
+              echo "TAG=$RAW_VERSION" >> $GITHUB_OUTPUT
+            fi
+  
+    create-rpm:
+      needs: gather-informations
+      uses: alumet-dev/packaging/.github/workflows/build_rpm.yaml@ci/create-rpm-package  # TODO CHANGE !!!
```
titouanj, line 76: Remember the TODO

.github/workflows/release.yaml
```diff
@@ -0,0 +1,142 @@
...
+              const latestRelease = await github.rest.repos.getLatestRelease({
+                owner: context.repo.owner,
+                repo: context.repo.repo
+              });
+  
+              // Retrieve informations from the latest release
+              const assets = latestRelease.data.assets.map(asset => asset.name);
+              let version = latestRelease.data.tag_name.replace(/^v/i, '');
+              let release = 1
+              if (assets.length === 0) {
```
titouanj, line 33: === 🤣🤮

.github/workflows/release.yaml
```diff
@@ -0,0 +1,142 @@
...
+                release_id: id
+              });
+  
+              for (const asset of assets.data) {
+                await github.rest.repos.deleteReleaseAsset({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  asset_id: asset.id
+                });
+              }
```
titouanj, line 131: Are we sure that we want to delete every asset?

titouanj (changes requested):

.github/workflows/release.yaml
AngeCyp: This seems to be a mistake :D

.github/workflows/release.yaml
AngeCyp: ❓"==" 🤣🤮
♻️
❓"===" 🤮🤣
🤙

.github/workflows/release.yaml
AngeCyp: Else, we can keep keep unwanted and old artifacts. Do you want me to change ?

titouanj: That's not what I meant. But we should replace the .rpm and .deb artifacts but not the others

AngeCyp: we should also replace container images. With this method, tar.gz and zip artifact are not deleted. Do you have an example of artifact we should keep ?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+              const firstAsset = assets[0]; // Retrieve first element of the list
+              const regex = /([0-9.]{3,})-([0-9]+)/;
+              const match = firstAsset.match(regex);
+              release = match ? match[2] : ''; // Retrieve the last release version using artifact name
+              if (release !== '') {
+                release = (parseInt(release) + 1).toString();
+              } else {
+                release = 1
+              }
+              version = match ? match[1] : '';
```
ahuret, line 45: looks like `version` will always be the same as the one calculated before

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+            RAW_VERSION=${{ github.event.release.tag_name }}
+            CLEAN_VERSION=${RAW_VERSION#v}
+            echo "VERSION=$CLEAN_VERSION" >> $GITHUB_OUTPUT
+            echo "RELEASE=1" >> $GITHUB_OUTPUT
+            echo "TAG=$RAW_VERSION" >> $GITHUB_OUTPUT
+          fi
+
+  create-rpm:
+    needs: gather-informations
+    uses: alumet-dev/packaging/.github/workflows/build_rpm.yaml@ci/create-rpm-package  # TODO CHANGE !!!
```
ahuret, line 76: we can change it now

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+            echo "VERSION=$CLEAN_VERSION" >> $GITHUB_OUTPUT
+            echo "RELEASE=${{ steps.get_latest_release.outputs.release }}" >> $GITHUB_OUTPUT
+            echo "TAG=${{ steps.get_latest_release.outputs.tag }}" >> $GITHUB_OUTPUT
+          else
+            RAW_VERSION=${{ github.event.release.tag_name }}
+            CLEAN_VERSION=${RAW_VERSION#v}
+            echo "VERSION=$CLEAN_VERSION" >> $GITHUB_OUTPUT
+            echo "RELEASE=1" >> $GITHUB_OUTPUT
+            echo "TAG=$RAW_VERSION" >> $GITHUB_OUTPUT
+          fi
```
ahuret, line 72: Proposed some refactor:

```suggestion
    - name: Get release, tag and version
      id: get_release_tag_version
      run: |
        if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
          RELEASE=${{ steps.get_latest_release.outputs.release }}
          TAG=${{ steps.get_latest_release.outputs.tag }}
          if [ -z ${{ steps.get_latest_release.outputs.version }} ]; then 
            VERSION=${TAG#v}
          else 
            VERSION=${{ steps.get_latest_release.outputs.version }}
          fi
        else
          RELEASE=1
          TAG=${{ github.event.release.tag_name }}
          VERSION=${TAG#v}
        fi
        echo "VERSION=$VERSION" >> $GITHUB_OUTPUT
        echo "RELEASE=$RELEASE" >> $GITHUB_OUTPUT
        echo "TAG=$TAG" >> $GITHUB_OUTPUT
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Download RPM artifact
+        uses: actions/download-artifact@v4
+        with:
+          pattern: '**/*.rpm'
+          path: ./rpm
+
+      - name: Release
```
ahuret, line 98: "Create new release"

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+                repo: context.repo.repo,
+                asset_id: asset.id
+              });
+            }
+            console.log(`Release id: ${id}`);
+            console.log(`Release tag: ${tag}`);
+            core.setOutput("release_id", id);
+            core.setOutput("release_tag", tag);
+
+      - name: Upload new artifacts to github release
```
ahuret, line 137: "Upload RPM artifacts to github release"
cause we won't use this step it for every artifacts

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+            for (const asset of assets.data) {
+              await github.rest.repos.deleteReleaseAsset({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                asset_id: asset.id
+              });
+            }
+            console.log(`Release id: ${id}`);
+            console.log(`Release tag: ${tag}`);
+            core.setOutput("release_id", id);
```
ahuret, line 134: it doesn't seems to be used

.github/workflows/release.yaml
ahuret: also this step is always executed but use `get_latest_release_and_delete_old_artifacts.release_tag` output how does it works in case ` github.event_name != 'workflow_dispatch'` ?
maybe you should add an if parameter to make explicit that it's the update release case.

ahuret (commented):
Added few comments
I believe the steps would be clearer if there were more context regarding the state of the release (whether it’s creating or updating). This would make the workflow easier to follow. It might help to include suffixes like ‘(Create case)’ and ‘(Update case)’ in the step titles, to clearly indicate which state each step is related to.

.github/workflows/release.yaml
AngeCyp: artefact from a previous time

.github/workflows/release.yaml
```diff
@@ -0,0 +1,138 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo
+            });
+
+            // Retrieve informations from the latest release
+            const assets = latestRelease.data.assets.map(asset => asset.name);
+            let version = latestRelease.data.tag_name.replace(/^v/i, '');
+            let release = 1
+            if (assets.length === 0) {
+              console.log("No artifacts attached to the latest release. Using 1 as release number.");
```
titouanj, line 34: Use core.notice?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,138 @@
...
+            TAG=${{ github.event.release.tag_name }}
+            VERSION=${TAG#v}
+          fi
+          echo "VERSION=$VERSION" >> $GITHUB_OUTPUT
+          echo "RELEASE=$RELEASE" >> $GITHUB_OUTPUT
+          echo "TAG=$TAG" >> $GITHUB_OUTPUT
+
+  create-rpm:
+    needs: gather-informations
+    uses: alumet-dev/packaging/.github/workflows/build_rpm.yaml@main  # TODO CHANGE !!!
```
titouanj, line 74: TODO

.github/workflows/release.yaml
```diff
@@ -0,0 +1,144 @@
...
+          else
+            RELEASE=1
+            TAG=${{ github.event.release.tag_name }}
+            VERSION=${TAG#v}
+          fi
+          echo "VERSION=$VERSION" >> $GITHUB_OUTPUT
+          echo "RELEASE=$RELEASE" >> $GITHUB_OUTPUT
+          echo "TAG=$TAG" >> $GITHUB_OUTPUT
+
+  create-rpm:
```
ahuret, line 72: let's call it `build-rpm` no ? to follow same name as in alumet-dev/packaging

.github/workflows/release.yaml
```diff
@@ -0,0 +1,144 @@
...
+          fi
+          echo "VERSION=$VERSION" >> $GITHUB_OUTPUT
+          echo "RELEASE=$RELEASE" >> $GITHUB_OUTPUT
+          echo "TAG=$TAG" >> $GITHUB_OUTPUT
+
+  build-rpm:
+    needs: gather-informations
+    uses: alumet-dev/packaging/.github/workflows/build_rpm.yaml@main
+    with:
+      target-architecture: x86_64
```
ahuret, line 76: use `arch` instead

.github/workflows/release.yaml
```diff
@@ -0,0 +1,144 @@
...
+      build-version: ${{ needs.gather-informations.outputs.version }}
```
ahuret, line 77: use `version` instead

.github/workflows/release.yaml
```diff
@@ -0,0 +1,144 @@
...
+
+            for (const asset of assets.data) {
+              try {
+                if ( asset.name.endsWith(".deb") || asset.name.endsWith(".rpm")) {
+                  await github.rest.repos.deleteReleaseAsset({
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    asset_id: asset.id
+                  });
+                } catch (error) {
```
ahuret, line 131: it fails see error here https://github.com/ahuret/alumet/actions/runs/14039822895/job/39307258080#step:5:51

TheElectronWill:
Can you build Deb packages too?

AngeCyp:
@TheElectronWill Done

.github/workflows/release.yaml
```diff
@@ -0,0 +1,158 @@
...
+            // Retrieve the latest release
+            const latestRelease = await github.rest.repos.getLatestRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo
+            });
+
+            // Retrieve informations from the latest release
+            const assets = latestRelease.data.assets.map(asset => asset.name);
+            let version = latestRelease.data.tag_name.replace(/^v/i, '');
+            let release = 1
```
titouanj, line 32:
```suggestion
            let release = 1;
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,158 @@
...
+              core.notice("No artifacts attached to the latest release. Using 1 as release number.");
+            } else {
+              const firstAsset = assets[0]; // Retrieve first element of the list
+              const regex = /([0-9.]{3,})-([0-9]+)/;
+              const match = firstAsset.match(regex);
+              release = match ? match[2] : ''; // Retrieve the last release version using artifact name
+              if (release !== '') {
+                release = (parseInt(release) + 1).toString();
+              } else {
+                release = 1
```
titouanj, line 43:
```suggestion
                release = 1;
```

titouanj (changes requested):
JS is really a weird thing

ahuret (approved):

titouanj (approved):

PR #138, opened 2025-03-31
feat(tests): unit tests + integration tests

ahuret:
- added unit tests
- introduced integration tests using new tests API
- reached 80%+ code tests coverage
- some small changes (doc+refactor of current influxdb code)


TheElectronWill:
Commit messages should begin with `test(plugin-influxdb)`

plugin-influxdb/src/lib.rs
```diff
@@ -174,22 +161,60 @@ impl Output for InfluxDbOutput {
...
+        }
+        AttributeAs::Field => {
+            // default is field => tag only if in set
+            attributes_as_tags.contains(key)
+        }
+    }
+}
+
+// Check if the tag key is reserved and in case it is append a prefix
+fn ensure_reserved_tag_key(reserved_tags: [&str; 4], tag_key: String) -> String {
```
TheElectronWill, line 185: Since the list of reserved tags is constant and depends on InfluxDB's spec, I think it would make more sense to use that constant in the function instead of passing it as a parameter.

plugin-influxdb/src/lib.rs
```diff
@@ -174,22 +161,60 @@ impl Output for InfluxDbOutput {
...
+    if reserved_tags.contains(&tag_key.as_str()) {
+        format!("alumet_attribute__{tag_key}")
+    } else {
+        tag_key
+    }
+}
+
+// Check if the field key is reserved and in case it is return a predefined key
+fn ensure_reserved_field_key<'a>(reserved_field: &'a str, field_key: &'a str) -> &'a str {
```
TheElectronWill, line 194: Similar comment here.

plugin-influxdb/src/lib.rs
TheElectronWill: How about calling the function `ensure_valid_tag_key` instead? It tells more what it does: ensure that it's valid (not "ensure that it's reserved). (please don't hurt me for starting a naming war :smile:)

plugin-influxdb/src/lib.rs
```diff
@@ -174,22 +161,60 @@ impl Output for InfluxDbOutput {
...
+    pub token: String,
+    pub org: String,
+    pub bucket: String,
+    pub attributes_as: AttributeAs,
+    pub attributes_as_tags: Option<HashSet<String>>,
+    pub attributes_as_fields: Option<HashSet<String>>,
 }
 
 /// How to serialize Alumet attributes by default?
 #[derive(Serialize, Deserialize)]
```
TheElectronWill, line 215:
```suggestion
#[derive(Serialize, Deserialize, Clone, Copy)]
```
Making the `AttributeAs` enum `Copy` will simplify its use: you will be able to remove all the references everywhere and just pass it directly. It makes sense because the enum is very simple (only two plain values), so it will end up as one byte, so there's no need to pass references everywhere.

For instance you'll be able to do:
```rs
assert_eq!(partition_tag("is_a_tag", AttributeAs::Tag, &tags, &fields), true);
```

TheElectronWill (changes requested):
Looks good overall, I have some "quality improvement" remarks.

plugin-influxdb/src/influxdb2.rs
```diff
@@ -262,10 +268,104 @@ mod tests {
...
+    async fn write() {
+        let mut server = Server::new_async().await;
+
+        let token = "sometoken";
+        let org = "someorg";
+        let bucket = "somebucket";
+
+        let influx_client = Client::new(server.url(), String::from(token));
+
+        for tested_line in get_tested_lines() {
```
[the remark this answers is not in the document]
ahuret: It's needed to make different tests in an uniform way.
Here we test two different cases without having to re-code the whole testing logic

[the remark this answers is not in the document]
titouanj: I understand why you added this function, but my question was more about the fact that you're calling it multiple times to get the exact same result. But since it is really fast, I don't care that much. I was more curious

plugin-influxdb/tests/influxdb.rs
```diff
@@ -0,0 +1,118 @@
...
+    use crate::fakeplugin::TestsPlugin;
+    use alumet::agent::plugin::PluginInfo;
+    use alumet::plugin::PluginMetadata;
+    use alumet::{
+        agent::{self, plugin::PluginSet},
+        measurement::{MeasurementBuffer, MeasurementPoint, WrappedMeasurementValue},
+        pipeline::naming::OutputName,
+        resources::{Resource, ResourceConsumer},
+        test::{runtime::OutputCheckInputContext, RuntimeExpectations},
+    };
```
titouanj, line 22: I think this should be formatted

plugin-influxdb/src/influxdb2.rs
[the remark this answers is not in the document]
ahuret: ah now got your point ^^ actually because cannot use some intermediate variable to store it (it's neither const or static)

AngeCyp (commented):
Seems good IMO

AngeCyp (approved):

TheElectronWill (approved):

PR #139, opened 2025-03-31
Make metric registration more configurable

### LINKED ISSUE #137: Accept compatible metric definitions by default
It would be useful to make `alumet.create_metric` more lenient: accept multiple calls to `create_metric` with the same metric.

TheElectronWill:
New options to choose (not always available in public plugin API, on purpose):
- `DuplicateCriteria`: defines the rules that make a new metric a duplicate.
- `DuplicateReaction`: what to do when there is a duplicate

Resolves #137.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,6 +51,7 @@ dependencies = [
  "indoc",
  "libc",
  "log",
+ "num_enum",
  "ordered-float",
  "serde",
  "serial_test",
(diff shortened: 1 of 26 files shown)


TheElectronWill:
The PR grew a bit more because I've just found the root cause of a flaky test: a bug in the testing module. Fixing the bug required new features in the core of Alumet.

PR #144, opened 2025-04-07
Push interval in opentelemetry plugin

guigomcha:
This way the user can configure how often (in secs) the metrics are pushed to the OTEL collector.


guigomcha:
@TheElectronWill  ready here

TheElectronWill (approved):

guigomcha:
@TheElectronWill CI fails for a timeout in the runner... could you rerun it? I don't seem to have account to do it

guigomcha:
@TheElectronWill It seems that I still cannot merge ...

PR #152, opened 2025-04-10
Refactor cgroupv2 plugin - metrics harmonization

ahuret:
This PR aims to harmonize collected metrics with plugins cgroupv1 and procfs.

In that direction we kept only one cpu related metric called `cpu_time_delta` and gather old cgroup_cpu_usage_total, cgroup_cpu_usage_user and cgroup_cpu_usage_system that are still collected but as a single metric with different `kind` attributes (total, usage and system).

Also added a new metric `memory_usage`.
This metric is sourced from memory.current file from cgroupv2.
It represents the resident (RSS) memory (kind attribute), that makes it consistend with metrics collected in cgroupv1 and procfs.
Removed the metric cgroup_memory_total that was calculated based on other cgroup_memory measurements, since it's recommended to collect memory.current to have the total memory currently used by a process (a bit like top does for example), it didn't make sense anymore to try to compute it on our side.

Made some renaming especially on memory.stat related variables as now we collect two different files related to metrics it make sense to explicit it.

### DIFF
--- a/plugin-cgroupv2/src/cgroupv2.rs
+++ b/plugin-cgroupv2/src/cgroupv2.rs
@@ -24,14 +24,26 @@ pub struct CgroupMeasurements {
     pub cpu_time_user_mode: u64,
     /// CPU in system mode usage time by the cgroup.
     pub cpu_time_system_mode: u64,
+    /// Resident memory usage (RSS) currently used by the cgroup.
+    pub memory_usage_resident: u64,
     /// Anonymous used memory, corresponding to running process and various allocated memory.
     pub memory_anonymous: u64,
-    // Files memory, corresponding to open files and descriptors.
+    /// Files memory, corresponding to open files and descriptors.
     pub memory_file: u64,
-    // Memory reserved for kernel operations.
+    /// Memory reserved for kernel operations.
     pub memory_kernel: u64,
     /// Memory used to manage correspondence between virtual and physical addresses.
     pub memory_pagetables: u64,
+}
+
+impl CgroupMeasurements {
+    pub fn load_memory_current_from_str(&mut self, s: &str) -> Result<()> {
+        self.memory_usage_resident = s
+            .trim()
+            .parse::<u64>()
+            .map_err(|e| anyhow::anyhow!("Failed to parse '{}': {}", s, e))?;
+        Ok(())
+    }
 }
 
 impl FromStr for CgroupMeasurements {
@@ -46,6 +58,7 @@ impl FromStr for CgroupMeasurements {
             cpu_time_total: 0,
             cpu_time_user_mode: 0,
             cpu_time_system_mode: 0,
+            memory_usage_resident: 0,
             memory_anonymous: 0,
             memory_file: 0,
             memory_kernel: 0,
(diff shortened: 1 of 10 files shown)


plugin-cgroupv2/src/k8s/probe.rs
```diff
@@ -168,7 +181,7 @@ impl alumet::pipeline::Source for K8SProbe {
         let m_tot = create_measurement_point(
             timestamp,
             self.memory_total,
-            self.cgroup_v2_metric_file.consumer_memory.clone(),
+            self.cgroup_v2_metric_file.consumer_memory_stat.clone(),
             mem_total_value,
             &metrics,
         );
```
titouanj, line 187: Since memory.current is "The total amount of memory currently being used by the cgroup and its descendants." [source](https://docs.kernel.org/admin-guide/cgroup-v2.html). Does it still make sense to try to compute the same value but differently?

ahuret: removed the memory_total metric !

plugin-cgroupv2/src/k8s/probe.rs
titouanj, line 1: The only things that differs from the poll of oar3 are the extra attributes specific to k8s. I think we can factorize this code

TheElectronWill:
I've noticed that the tests of the cgroupv2 plugin depends on an outdated version of mockito, which is out of sync with other plugins such as influxdb. Can you update the dev dependency from `mockito = "0.31"` to `mockito = "1.7.0"`?

The `Token` structure seems a bit weird to me:
```rs
/// Kubernetes token and way to retrieve it.
#[derive(Debug)]
pub struct Token {
    /// Way to obtain a token, in file or with Kubectl command.
    retrieval: TokenRetrieval,
    /// Thread safe token data.
    data: Arc<RwLock<TokenData>>,
    /// Public path for a token if we store it in a file.
    pub path: Option<String>,
}
```

If `TokenRetrieval` is `File`, it looks like the `path` must always be set. This is error-prone because there is no compile-time guarantee: `retrieval` should be either `Kubectl` or `File(path_to_file)` to avoid that. It turns out that, if `TokenRetrieval` is `File` and `path` is `None`, a default path is used. IMO this is quite confusing.

The proper fix would be to remove `path`, store the file path in the enum, and provide a proper constructor for the "default file" case, e.g. `Token::with_default_file()`.

plugin-cgroupv2/src/k8s/probe.rs
ahuret: sure, there are several things to refactor to make cgroupv2 common code bigger. Should it be done as part of this PR ?

TheElectronWill (changes requested):
Additional remark:
many unit tests create multiple nested paths like this:
```rs
let tmp = tempdir().unwrap();
let root = tmp.path().join("test-alumet-plugin-k8s/var_13/");

if root.exists() {
    std::fs::remove_dir_all(&root).unwrap();
}

let dir = root.join("run/secrets/kubernetes.io/serviceaccount/");
std::fs::create_dir_all(&dir).unwrap();

let path = dir.join("token_13");
```

I think that it comes from the first version of the tests, which didn't use `tempfile`. With `tempfile::tempdir`, it's useless to use the name of the test in the path (because it returns a new directory with a unique randomly-generated name, and removes it when the test ends). The whole thing could be simplified to keep only a simple token file directly in the temp dir.

plugin-cgroupv2/src/cgroupv2.rs
```diff
@@ -24,16 +24,30 @@ pub struct CgroupMeasurements {
...
     /// Memory used to manage correspondence between virtual and physical addresses.
     pub memory_pagetables: u64,
 }
 
+impl CgroupMeasurements {
+    pub fn load_memory_current_from_str(&mut self, s: &str) -> Result<()> {
+        match s.trim().parse::<u64>() {
+            Ok(value) => self.memory_usage_resident = value,
+            Err(e) => {
+                eprintln!("Failed to parse '{}': {}", s, e);
```
TheElectronWill, line 44: You should return an error instead of printing to stderr here :p

TheElectronWill (changes requested):

plugin-cgroupv2/src/cgroupv2.rs
ahuret: resolved !

TheElectronWill (dismissed):

TheElectronWill:
A little rebase and it'll be good

PR #158, opened 2025-04-15
Quick 0.8.1 release

TheElectronWill:
The difference with v0.8.0 is that I've enabled aggregation, energy-attribution and energy-estimation-tdp in alumet-agent.

### DIFF
--- a/agent/src/bin/main.rs
+++ b/agent/src/bin/main.rs
@@ -30,6 +30,9 @@ fn load_plugins_metadata() -> Vec<PluginMetadata> {
         plugin_relay::client::RelayClientPlugin,
         plugin_relay::server::RelayServerPlugin,
         plugin_opentelemetry::OpenTelemetryPlugin,
+        plugin_aggregation::AggregationPlugin,
+        plugin_energy_attribution::EnergyAttributionPlugin,
+        plugin_energy_estimation_tdp::EnergyEstimationTdpPlugin,
     ];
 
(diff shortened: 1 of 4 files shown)


titouanj (approved):

agent/src/bin/main.rs
```diff
@@ -30,6 +30,9 @@ fn load_plugins_metadata() -> Vec<PluginMetadata> {
         plugin_relay::client::RelayClientPlugin,
         plugin_relay::server::RelayServerPlugin,
         plugin_opentelemetry::OpenTelemetryPlugin,
+        plugin_aggregation::AggregationPlugin,
+        plugin_energy_attribution::EnergyAttributionPlugin,
+        plugin_energy_estimation_tdp::EnergyEstimationTdpPlugin,
```
AngeCyp, line 35: Not in alphabetical order :/

AngeCyp (changes requested):

agent/src/bin/main.rs
TheElectronWill: The order does mean something here. For example, if aggregation is before attribution, the aggregation transforms will be applied before the attribution transforms by default.

AngeCyp: oh okay, so my bad. Thanks for your response :D

AngeCyp: I let you resolve this conversation

PR #161, opened 2025-04-17
ci: Ensure that the git history of each PR branch is linear

TheElectronWill:
A new action enforces the model "rebase, then merge" where each feature branch has a linear history, but can be merged into master with a merge commit. If the PR branch contains a merge commit, the check will fail.

### DIFF
--- a/.github/workflows/check_doc.yaml
+++ b/.github/workflows/check_doc.yaml
@@ -1,4 +1,3 @@
----
 name: Check documentation
 on: [pull_request, push]
 
@@ -8,7 +7,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 0
+          fetch-depth: 1
       - name: Spell check everything
         uses: streetsidesoftware/cspell-action@v6
         with:
--- a/.github/workflows/check_git.yaml
+++ b/.github/workflows/check_git.yaml
@@ -1,0 +1,21 @@
+name: Ensure a clean Git history
+on: [pull_request]
+
+jobs:
+  git-checkup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Check that the Git history is linear
+        run: |
+          NB_MERGE_COMMITS_ON_BRANCH=$(git rev-list --merges --count ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }})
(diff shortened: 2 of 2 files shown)


TheElectronWill:
Example: PR #162

.github/workflows/check_doc.yaml
```diff
@@ -1,4 +1,3 @@
----
```
titouanj, line 1: 😢

.github/workflows/check_git.yaml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Check that the Git history is linear
+        run: |
+          NB_MERGE_COMMITS_ON_BRANCH=$(git rev-list --min-parents=2 --count ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }})
```
titouanj, line 16: To make it clearer 
```suggestion
          NB_MERGE_COMMITS_ON_BRANCH=$(git rev-list --merges --count ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }})
```

.github/workflows/check_git.yaml
```diff
@@ -0,0 +1,21 @@
...
+          if [ $NB_MERGE_COMMITS_ON_BRANCH -ne 0 ]; then
+            echo "ERROR: The git history of your branch is non-linear: $NB_MERGE_COMMITS_ON_BRANCH merge commits detected"
+            echo "When updating your branch, you must rebase instead of merging."
```
titouanj, line 19: Use error workflow command: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message

For example:
```suggestion
            echo "::errorr::The git history of your branch is non-linear: $NB_MERGE_COMMITS_ON_BRANCH merge commits detected"
            echo "::error::When updating your branch, you must rebase instead of merging."
```

titouanj (changes requested):

.github/workflows/check_doc.yaml
TheElectronWill: ah yes
Why is there a `---`? This is not a stream of documents.

titouanj: That's just the start of a document according to the yaml spec

titouanj (approved):

.github/workflows/check_doc.yaml
titouanj: Okay I get your point

AngeCyp (approved):

Issue #159, opened 2025-04-16
Refactor plugin cgroupv2 - code cleanup


TheElectronWill:
See my comments on PR #152.

TODO:
- [x] update dev dependencies
- [x] simplify tests by leveraging tempfile::tempdir
- [x] make Token structure more consistent and type safe
- [x] factorize common code when it's good to do so

ahuret:
update dev dependencies => mockito issue => done in https://github.com/alumet-dev/alumet/pull/152

ahuret:
simplify tests by levaraging tempfile => https://github.com/alumet-dev/alumet/pull/165
make Token structure more consistent => https://github.com/alumet-dev/alumet/pull/164

TheElectronWill:
Done in #226

PR #165, opened 2025-04-18
refactor(plugin-cgroupv2): k8s submodule - rework tempfiles in tests

ahuret:
This PR fix the tempfile related part of https://github.com/alumet-dev/alumet/issues/159

### DIFF
--- a/plugin-cgroupv2/src/k8s/token.rs
+++ b/plugin-cgroupv2/src/k8s/token.rs
@@ -245,12 +245,7 @@ mod tests {
     // Test `refresh` function with valid file and token
     #[tokio::test]
     async fn test_refresh_with_valid_token() {
-        let tmp = tempdir().unwrap();
-        let root = tmp.path().join("test-alumet-plugin-k8s/var_1/");
-
-        if root.exists() {
-            std::fs::remove_dir_all(&root).unwrap();
-        }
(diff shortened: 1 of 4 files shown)


titouanj (commented):
need rebase

plugin-cgroupv2/src/k8s/token.rs
```diff
@@ -245,12 +245,7 @@ mod tests {
     // Test `refresh` function with valid file and token
     #[tokio::test]
     async fn test_refresh_with_valid_token() {
-        let tmp = tempdir().unwrap();
-        let root = tmp.path().join("test-alumet-plugin-k8s/var_1/");
-
-        if root.exists() {
-            std::fs::remove_dir_all(&root).unwrap();
-        }
+        let root = tempdir().unwrap().into_path();
```
TheElectronWill, line 248: Beware, `into_path` consumes the TempDir and disables auto-cleaning. It has been deprecated and renamed to `keep` in recent versions of `tempfile`. If we want to clean the directories after the test (which I think is a good idea), we should keep the tempdir on its own and get its path with `path`, in another variable.

TheElectronWill (changes requested):

ahuret:
closing this PR cause changed will be made in https://github.com/alumet-dev/alumet/pull/172

PR #166, opened 2025-04-18
New plugin to send measurements to OpenSearch/ElasticSearch directly

TheElectronWill:
The OpenTelemetry plugin is not enough, because it requires an intermediary program to translate between the otel protocol and the OpenSearch/ElasticSearch REST API.

### DIFF
--- a/plugin-elasticsearch/src/api/client.rs
+++ b/plugin-elasticsearch/src/api/client.rs
@@ -1,0 +1,173 @@
+use std::{collections::HashMap, fmt::Debug};
+
+use alumet::{
+    measurement::MeasurementBuffer,
+    pipeline::{control::request, elements::output::OutputContext},
+};
+use anyhow::Context;
+use reqwest::{
+    header::{HeaderMap, HeaderValue, InvalidHeaderValue},
+    Url,
+};
+
+use super::ser::{CreateIndexTemplate, IndexTemplate, Serializer};
+
+// OpenSearch/ElasticSearch API client.
+pub struct Client {
+    serializer: Serializer,
+    client: reqwest::blocking::Client,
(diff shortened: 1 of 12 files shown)


plugin-elasticsearch/src/api/client.rs
```diff
@@ -0,0 +1,173 @@
...
+        let request = self.client.put(url).body(body).header(
+            reqwest::header::CONTENT_TYPE,
+            HeaderValue::from_static("application/x-ndjson"),
+        );
+        log::trace!("sending {request:?}");
+        let response = request.send().context("could not send request")?;
+
+        log::trace!("got response {response:?}");
+        Self::handle_response(response)?;
+        // let failed_items = response
```
ahuret, line 121: let's remove comment ?

plugin-elasticsearch/src/api/ser.rs
```diff
@@ -0,0 +1,157 @@
...
+            (
+                field.to_string(),
+                json!({
+                    "type": data_type
+                }),
+            )
+        }
+
+        let index_props = serde_json::Map::from_iter([
+            field_with_type("@timestamp", "date_nanos"),
```
ahuret, line 34: maybe this behavior of defining fields and types should be placed as a `DocMeasurement` method, as it's his own specific knowledge (note that `DocMeasurement` mentions about the columns names in the `serialize` method).

ahuret (changes requested):
Added 2 comments and otherwise LGTM

plugin-elasticsearch/src/api/client.rs
TheElectronWill: oops! Yes it should definitely be removed

PR #169, opened 2025-04-23
refactor(plugin-cgroupv1): factorized code in prevision for slurm submodule

ahuret:
This PR add some factorization and cleanup in the plugin-cgroupv1 plugin.
It's needed in prevention of the Slurm submodule (that also use cgroupv1).

There's a new `Cgroupv1Probe` that will manage collection of measurements.
`OAR2Probe` use it and add some higher level things specific to OAR2.

Also refactored the OAR2 internals to not do things twice (refactored source creation).

Meanwhile I did some renaming and cleanups.


plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -1,34 +1,38 @@
...
 impl Metrics {
     pub fn new(alumet: &mut AlumetPluginStart) -> Result<Self, MetricCreationError> {
         let nsec = PrefixedUnit::nano(Unit::Second);
 
         Ok(Self {
-            cpu_metric: alumet.create_metric::<u64>(
+            cpu_time_delta: alumet.create_metric::<u64>(
                 "cpu_time_delta",
                 nsec,
                 "Total CPU time consumed by the cgroup (in nanoseconds).",
```
AngeCyp, line 33:
```suggestion
                "Total CPU time consumed by the cgroup.",
```
As the unit is already given in the metric (it's what was done on cgroupv2)

plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -1,34 +1,38 @@
...
             )?,
-            memory_metric: alumet.create_metric::<u64>(
+            memory_usage: alumet.create_metric::<u64>(
                 "memory_usage",
                 Unit::Byte,
                 "Total memory usage by the cgroup (in bytes).",
```
AngeCyp, line 38:
```suggestion
                "Total memory usage by the cgroup.",
```
As the unit is already given in the metric (it's what was done on cgroupv2)

plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -37,47 +41,100 @@ impl Metrics {
...
+        }
+        Ok(probe)
+    }
+
+    pub fn collect_measurements(
+        &mut self,
+        timestamp: Timestamp,
+        additional_attrs: &HashMap<String, String>,
+    ) -> Result<Vec<MeasurementPoint>, PollError> {
+        let mut measurement_points: Vec<MeasurementPoint> = Vec::new();
```
AngeCyp, line 90:
```suggestion
        let mut measurement_points = Vec::<MeasurementPoint>::new();
```
IDK if it is the best but I remember this kind of trick to not always display on the left side the type of variable

AngeCyp: Maybe don't change your code

plugin-cgroupv1/src/oar2/probe.rs
```diff
@@ -1,26 +1,39 @@
...
-use crate::cgroupv1::gather_value;
+use crate::cgroupv1::{Cgroupv1Probe, Metrics};
 
 pub struct OAR2Probe {
-    pub cpu_metric_counter_diff: CounterDiff,
-    pub cpu_metric: TypedMetricId<u64>,
-    pub memory_metric: TypedMetricId<u64>,
-    pub oar2_metric_file: OpenedCgroupv1,
+    cgroupv1: Cgroupv1Probe,
+    additional_attrs: HashMap<String, String>,
```
AngeCyp, line 11: So the jobID is lost between additional attributs ? 😲
We can discuss about it

AngeCyp (changes requested):

plugin-cgroupv1/src/cgroupv1.rs
ahuret: imo both are equivalent in term of readibility but I will change it as you said .

plugin-cgroupv1/src/oar2/probe.rs
ahuret: yes cause cgroupv1 actually don't have a concept of "job". So even if submodules OAR2 and Slurm both have this concept, I think it doesn't make sense to add this semantic in cgroupv1 related code.

AngeCyp: We already are in the OAR2 submodules,  so I don't wanted to add the jobID in the crgoupv1 struct, I was more in  a way to add an other field in the struct of the oar2 which contain a jobID and an other one containing all other attributes such as:
```suggestion
    cgroupv1: Cgroupv1Probe,
    jobID: u64
    additional_attrs: HashMap<String, String>,
```

AngeCyp (approved):
LGTM, just maybe one last discussion nor completed. Still to discuss maybe w/ the architect

plugin-cgroupv1/src/oar2/probe.rs
ahuret: ah yes sorry! :P 
hum in fact there's no need to store jobID as a field here since it's registered in the additional_attrs at the creation of the OAR2Probe.

AngeCyp (dismissed):

plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -1,83 +1,140 @@
...
+                CounterDiffUpdate::FirstTime => None,
+                CounterDiffUpdate::Difference(diff) => Some(diff),
+                CounterDiffUpdate::CorrectedDifference(diff) => Some(diff),
+            };
+            if let Some(cpu_time_delta_value) = cpu_time_delta {
+                measurement_points.push(MeasurementPoint::new(
+                    timestamp,
+                    self.metrics.cpu_time_delta,
+                    Resource::LocalMachine,
+                    self.cpu_time_delta_consumer.clone().unwrap(),
```
titouanj, line 113: You can add a context error to explain that the consumer is missing or something like that

plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -1,83 +1,140 @@
...
+
+        if let Some(cpu_time_delta_file) = &mut self.cpu_time_delta_file {
+            buffer.clear();
+            cpu_time_delta_file.rewind()?;
+            cpu_time_delta_file.read_to_string(&mut buffer)?;
+            let cpu_time_total = buffer.trim().parse::<u64>()?;
+            let cpu_time_delta = match self
+                .cpu_time_delta_counter_diff
+                .as_mut()
+                .unwrap()
```
titouanj, line 101: Do we really want to have a panic here?

plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -1,83 +1,140 @@
...
-    }
+                self.memory_usage_consumer.clone().unwrap(),
+                memory_usage_u64,
+            ));
+        }
+        for (key, value) in additional_attrs {
+            for point in &mut measurement_points {
+                point.add_attr(key.to_string(), value.to_string());
+            }
+        }
```
titouanj, line 136: You can add the attributes to the point when you create it with .with_attr, .with_attr_vec or .with_attr_map

```rust
let new_m = MeasurementPoint::new().with_attr_vec(point_attributes);
```

plugin-cgroupv1/src/oar2/probe.rs
titouanj, line 11: You might want to change this into a vec to be able to use .with_attr_vec (like mentioned above) else, I think you'll be the first one to use the .with_attr_map

AngeCyp (commented):
LGTM but still waiting for an other approval

titouanj (approved):

PR #170, opened 2025-04-25
Refactor/plugin cgroupv1 add tests

ahuret:
This PR brings some unit tests and integration tests in plugin-cgroupv1.
Code covered by tests is now up to 92% !
It contains:
- Mock OAR Cgroup files for different submitters and different jobs
- Ensure Startup of plugin is fine, metrics correctly registered and expected source started
- Ensure Runtime of plugin is fine, first polls and subsequents return correct measurements, consumers paths and attributes. Also ensure new cgroup that pop up at runtime will get measured.

:warning:  Note that this PR is based on top of https://github.com/alumet-dev/alumet/pull/169 so we must first close this PR.


AngeCyp:
Could you please update the branch ? @ahuret

plugin-cgroupv1/src/oar2/plugin.rs
```diff
@@ -134,131 +100,29 @@ impl AlumetPlugin for Oar2Plugin {
...
-            poll_interval,
+            metrics,
+            trigger,
             rt,
         };
         let mut watcher = notify::recommended_watcher(handler)?;
 
-        watcher.watch(&config_path.join("cpuacct/oar"), RecursiveMode::NonRecursive)?;
-        watcher.watch(&config_path.join("memory/oar"), RecursiveMode::NonRecursive)?;
+        println!("START WATCHER");
```
TheElectronWill, line 123: a little println is lost here, please bring it back to his parents and use `log` instead :)

plugin-cgroupv1/tests/oar2.rs
```diff
@@ -0,0 +1,292 @@
...
+                            job_id: "666",
+                            cgroup_files: vec![
+                                CgroupFile::CpuAcctUsage(123123123),
+                                CgroupFile::MemoryUsageInBytes(321321321),
+                            ],
+                        },
+                    ];
+                    if let Err(err) = create_oar_mocks(&path, mocks) {
+                        println!("error while creating oar mock: {err:?}");
+                        return;
```
TheElectronWill, line 175: Since it's in the tests you can just `.expect` on the result.

plugin-cgroupv1/tests/oar2.rs
```diff
@@ -0,0 +1,292 @@
...
+        agent.wait_for_shutdown(Duration::from_secs(10)).unwrap();
+
+        Ok(())
+    }
+
+    /// This test performs lot of different tests in order to verify the plugin run correctly at runtime:
+    /// It's orchestrated in a way to test different scenario: first poll and other polls, since first poll will not have measurements for CounterDiff metrics.
+    /// It also ensure cgroup that pop up during Alumet's runtime are correctly collected too.
+    #[test]
+    fn test_runtime() -> Result<(), Box<dyn std::error::Error>> {
```
TheElectronWill, line 93: Since you have used `unwrap` a lot, you can replace `?` by unwrap or expect and it'll be fine.
Another way would be to use `anyhow` Errors, which would be a bit more easy to write/read and would provide nice backtraces.

plugin-cgroupv1/src/oar2/probe.rs
```diff
@@ -1,26 +1,38 @@
...
 };
-use std::result::Result::Ok;
+use std::{collections::HashMap, result::Result::Ok};
 
-use super::utils::OpenedCgroupv1;
-
-use crate::cgroupv1::gather_value;
+use crate::cgroupv1::{Cgroupv1Probe, Metrics};
 
 pub struct OAR2Probe {
```
TheElectronWill, line 9:
```suggestion
pub struct Oar2Probe {
```
The struct should be renamed to adhere to Rust naming conventions.

plugin-cgroupv1/src/oar2/plugin.rs
```diff
@@ -270,6 +134,104 @@ impl AlumetPlugin for Oar2Plugin {
...
+        {
+            log::debug!("Paths: {paths:?}");
+            for path in paths {
+                if let Err(e) = handle_event_on_path(self, path.clone()) {
+                    log::error!("Unable to handle event on {}: {}", path.display(), e);
+                }
+            }
+        } else if let Err(e) = event {
+            log::error!("watch error: {:?}", e);
+        }
```
TheElectronWill, line 231: Style improvement: use a `match` here to clearly show the different possibilities: Ok with a kind of Create, Ok with other content, and Err.

plugin-cgroupv1/src/oar2/plugin.rs
```diff
@@ -270,6 +134,104 @@ impl AlumetPlugin for Oar2Plugin {
...
+                if job_name.chars().any(|c| c.is_numeric()) {
+                    let source = Oar2Plugin::new_job_source(
+                        &job_name,
+                        job_detect.metrics.clone(),
+                        &job_detect.cpuacct_controller_path,
+                        &job_detect.memory_controller_path,
+                    )?;
+
+                    let source_name = job_name;
+                    let new_job_source =
```
TheElectronWill, line 200: Optimization (maybe later): build a list of sources to create and use `request::create_many()`. You can put a TODO if you feel like it would be better in a follow up PR.

plugin-cgroupv1/src/cgroupv1.rs
```diff
@@ -148,3 +139,50 @@ impl Cgroupv1Probe {
...
     }
 }
+
+fn collect_memory_usage(buffer: &mut String, file: &mut File) -> Result<u64, anyhow::Error> {
+    collect_file_single_measurement(buffer, file)
+}
+
+fn collect_cpuacct_usage(buffer: &mut String, file: &mut File) -> Result<u64, anyhow::Error> {
+    collect_file_single_measurement(buffer, file)
+}
```
titouanj, line 149: I don't understand the need for these two wrappers

plugin-cgroupv1/tests/oar2.rs
```diff
@@ -0,0 +1,314 @@
...
+                    .join(cgroup_file.controller())
+                    .join("oar")
+                    .join(format!("{0}_{1}", mock.submitter, mock.job_id));
+                fs::create_dir_all(&cgroup_path)?;
+                cgroup_file.write(&cgroup_path)?;
+            }
+        }
+        Ok(())
+    }
+}
```
titouanj, line 314: Can you move this kind of code at the top of this file ? It makes the reviews/code comprehension much easier

plugin-cgroupv1/src/cgroupv1.rs
ahuret: `collect_memory_usage` and `collect_cpuacct_usage` both have knowledge on how to collect their respective purpose. It gives clear semantic. It's a good way to not let the parent caller assume about implementation details of specific collections.

For example we could later have a 3rd metric from a file having another format like `key=value`, the parent caller would call collect_3rd_metric in the same way than we do for cpuacct_usage and memory_usage.

TheElectronWill: I have to use the YAGNI card: "You Aren't Gonna Need It" :)
I agree that it's good to give clear semantics, but the distinction between memory and cpu is not in the collection function, it's in the probe.

plugin-cgroupv1/tests/oar2.rs
```diff
@@ -0,0 +1,314 @@
+#[cfg(test)]
```
TheElectronWill, line 1: There is no need to do this, every file in `tests` is automatically detected as a test :)

PR #187, opened 2025-05-19
feature(grace-hopper): add the new grace-hopper plugin with it's tests

AngeCyp:
Add the new plugin 'grace-hopper'. It's goal is to look at the directory where are stored files which expose energy consumption.

Sources: 
- https://docs.nvidia.com/grace-perf-tuning-guide/power-thermals.html


plugin-grace-hopper/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    fn start(&mut self, alumet: &mut alumet::plugin::AlumetPluginStart) -> anyhow::Result<()> {
+        let base_dir = self.config.root_path.to_string();
+        // Try to open the directory
+        if let Ok(entries) = fs::read_dir(base_dir) {
+            for entry in entries {
+                if let Ok(entry) = entry {
+                    let path = entry.path();
+                    // Check if it's a directory
+                    if path.is_dir() {
```
titouanj, line 52: I think that a fail fast would be more readable

plugin-grace-hopper/src/probe.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl Source for GraceHopperProbe {
+    fn poll(
+        &mut self,
+        measurements: &mut alumet::measurement::MeasurementAccumulator,
+        timestamp: alumet::measurement::Timestamp,
+    ) -> Result<(), alumet::pipeline::elements::error::PollError> {
+        let mut buffer = String::new();
```
ahuret, line 60: It make sense to decouple the raw collection value in another function. That make test easier.

In general I think having a first step of **collection of raw data** then a second step of **wrapping in Alumet's world** is cleaner and enhance testing.

plugin-grace-hopper/src/probe.rs
```diff
@@ -0,0 +1,88 @@
...
+        if let Some(file) = &mut self.file {
```
ahuret, line 61: I don't think this check make sense here, no file == no source. So a source should always have a file (remove option).

ahuret (changes requested):

titouanj (changes requested):

ahuret (changes requested):

plugin-grace-hopper/src/probe.rs
```diff
@@ -0,0 +1,136 @@
...
+pub struct GraceHopperProbe {
+    socket: String,
+    kind: String,
+    file: File,
+    consumer: ResourceConsumer,
+    metric: Option<TypedMetricId<u64>>,
+    _power_stats_interval: String,
+}
+
+impl GraceHopperProbe {
```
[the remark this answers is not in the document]
AngeCyp: Well thank you for this comment. We discussed a lot with @titouanj and @TheElectronWill and we decided to not split as you recommend the plugin. Actually the plugin is not large enough to justify this split and we will let the architecture as currently it is. Moreover regarding what is exposed by the files, this plugin should not be required to retrieve any more metrics.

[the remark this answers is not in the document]
ahuret: :+1:

titouanj (approved):

titouanj (dismissed):

titouanj (approved):

PR #188, opened 2025-05-19
bug(core): counter diff corrected difference overflow

ahuret:
This PR resolves an issue I had while doing tests using a CounterDiff: `"attempt to subtract with overflow`".

It happened cause of this block:

```rust
if new_value < prev {
    let diff = new_value - prev + self.max_value;
    CounterDiffUpdate::CorrectedDifference(diff)
}
```

In my case `new_value` was lower than `prev` and **since diff is a u64, it cannot be negative**.

Changed a little bit the formula to correct the behavior.

The new formula also contains a fix of a potential **off-by-one error** I suspect, by adding `+1`:

For example, if `max_value = 255`, `prev = 250`, and `new_value = 5`:
The counter went from 250 -> 251 -> 252 -> 253 -> 254 -> 255 -> 0 -> 1 -> 2 -> 3 -> 4 -> 5 .
That's a total difference of `11`.
Old formula: `5 - 250 + 255 = 10` .
New formula: `255 - 250 + 5 + 1 = 11` .

### DIFF
--- a/alumet/src/plugin/util.rs
+++ b/alumet/src/plugin/util.rs
@@ -31,6 +31,7 @@ pub struct CounterDiff {
 }
 
 /// Result of [`CounterDiff::update()`].
+#[derive(PartialEq, Debug)]
 pub enum CounterDiffUpdate {
     /// This is the first counter update, its value is not meaningful.
     FirstTime,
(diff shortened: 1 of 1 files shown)


AngeCyp:
Nice improvement, I have never faced this issue before.

TheElectronWill:
Oh, good catch!
Can you add a unit test for this?

TheElectronWill:
> Nice improvement, I have never faced this issue before.

The panic only happens in debug mode, that's probably why.

ahuret:
> Oh, good catch Antoine! Can you add a unit test for this?

I added unit tests!

titouanj (approved):

titouanj (approved):

PR #195, opened 2025-05-27
tests(plugin-rapl): refactored and added unit + integration tests

ahuret:
Added new unit tests to increase the tested code coverage (up to 70% now).
In order to do so I've done some refactoring, in particular:

- get a global consistency on how powercap and perf_event are initialized and the kind of objects they return
- removed old `PowerZoneHierarchy` that was actually not used in the plugin: now we just use the "flat" version, "top" was removed. This makes it consistent with powercap objects.
- simulate perf_event_open syscall using `nix` library in order to make unit testing of perf_event
- added new configs only available in tests to be able to pass path containing mocked data
- rewrite some functions that was not test-ready because of hardcoded constant paths.

Note that `perf_events` is only partially tested yet, we still miss integration tests for it. Biggest issue is about finding a way to mock syscall+file descriptor behavior while running it with alumet (unit testing part have been done yet).

### DIFF
--- a/plugin-rapl/Cargo.toml
+++ b/plugin-rapl/Cargo.toml
@@ -7,7 +7,6 @@ edition = "2021"
 
 [features]
 
-
 [dependencies]
 alumet = { path = "../alumet" }
 anyhow = "1.0.88"
@@ -17,6 +16,13 @@ log = "0.4.22"
 perf-event-open-sys2 = "5.0.6"
 regex = "1.10.6"
 serde = { version = "1.0.210", features = ["derive"] }
+once_cell = "1.21.3"
+
+[dev-dependencies]
+alumet = {path = "../alumet", features = ["test"]}
+toml = "0.8.19"
+tempfile = "3.20.0"
+nix = "0.30.1"
 
 [lints]
 workspace = true
--- a/plugin-rapl/src/lib.rs
+++ b/plugin-rapl/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{path::PathBuf, time::Duration};
+use std::time::Duration;
 
 use alumet::{
     pipeline::elements::source::{trigger, Source},
@@ -9,23 +9,50 @@ use alumet::{
     units::Unit,
 };
 use anyhow::{anyhow, Context};
-use indoc::indoc;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    consistency::{check_domains_consistency, SafeSubset},
-    perf_event::PerfEventProbe,
-    powercap::PowercapProbe,
+    consistency::{get_available_domains, SafeSubset},
+    perf_event::{PerfEventProbe, PowerEvent},
+    powercap::{PowerZone, PowercapProbe},
 };
+
+#[cfg(test)]
+use std::path::PathBuf;
 
 mod consistency;
 mod cpus;
 mod domains;
-mod perf_event;
+pub mod perf_event;
 mod powercap;
+
+#[cfg(test)]
+pub mod tests_mock;
 
 pub struct RaplPlugin {
     config: Config,
+}
+
+impl RaplPlugin {
+    #[cfg(not(test))]
+    fn get_all_power_events(&self) -> anyhow::Result<Vec<PowerEvent>> {
+        perf_event::all_power_events()
+    }
+
+    #[cfg(test)]
+    fn get_all_power_events(&self) -> anyhow::Result<Vec<PowerEvent>> {
+        perf_event::all_power_events_from_path(&self.config.perf_event_test_path)
+    }
+
+    #[cfg(not(test))]
+    fn get_all_power_zones(&self) -> anyhow::Result<Vec<PowerZone>> {
+        Ok(powercap::all_power_zones()?.flat)
+    }
+
+    #[cfg(test)]
+    fn get_all_power_zones(&self) -> anyhow::Result<Vec<PowerZone>> {
+        Ok(powercap::all_power_zones_from_path(&self.config.powercap_test_path)?.flat)
+    }
 }
 
 impl AlumetPlugin for RaplPlugin {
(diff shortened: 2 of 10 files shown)


agent/tests/local.rs
```diff
@@ -55,6 +55,8 @@ fn regen_config_with_plugins() -> anyhow::Result<()> {
         poll_interval = "1s"
         flush_interval = "5s"
         no_perf_events = false
+        perf_event_test_path = ""
+        powercap_test_path = ""
```
ahuret, line 59: this is a weird behavior, this is because the config regen test is done in a test context with --all-features and in this PR I added new configs only available for test-support feature.

plugin-rapl/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-
+default = []
+test-support = ["tempfile", "nix"]
```
TheElectronWill, line 10: these should be dev-dependencies, with the `test` cfg `#[cfg(test)]` instead of a custom feature

plugin-rapl/src/powercap.rs
```diff
@@ -17,24 +18,16 @@ use alumet::{
...
-
-const POWERCAP_RAPL_PATH: &str = "/sys/devices/virtual/powercap/intel-rapl";
+pub const POWERCAP_RAPL_PATH: &str = "/sys/devices/virtual/powercap/intel-rapl";
 const POWER_ZONE_PREFIX: &str = "intel-rapl";
 const POWERCAP_ENERGY_UNIT: f64 = 0.000_001; // 1 microJoules
 
 const PERMISSION_ADVICE: &str = "Try to adjust file permissions.";
 
-/// Hierarchy of power zones
-pub struct PowerZoneHierarchy {
```
TheElectronWill, line 29: Can you keep this one? I want to add a feature that shows the domain hierarchy, later.

plugin-rapl/src/lib.rs
```diff
@@ -9,25 +9,52 @@ use alumet::{
...
+impl RaplPlugin {
+    #[cfg(not(feature = "test-support"))]
+    fn get_all_power_events(&self) -> anyhow::Result<Vec<PowerEvent>> {
+        perf_event::all_power_events()
+    }
+
+    #[cfg(feature = "test-support")]
+    fn get_all_power_events(&self) -> anyhow::Result<Vec<PowerEvent>> {
+        perf_event::all_power_events_from_path(&self.config.perf_event_test_path)
+    }
```
TheElectronWill, line 45:
```suggestion
    fn get_all_power_events(&self) -> anyhow::Result<Vec<PowerEvent>> {
        #[cfg(not(test))]
        perf_event::all_power_events()
        
        #[cfg(test)]
        perf_event::all_power_events_from_path(&self.config.perf_event_test_path)
    }
```

plugin-rapl/tests/powercap.rs
```diff
@@ -0,0 +1,194 @@
+#[cfg(test)]
+mod tests {
```
TheElectronWill, line 2: Not needed, files in tests are automatically recognized as integration tests.

plugin-rapl/tests/powercap.rs
```diff
@@ -0,0 +1,194 @@
...
+            },
+        ];
+
+        create_mock_layout(base_path.clone(), &entries)?;
+
+        let source_config = Config {
+            poll_interval: Duration::from_secs(1),
+            flush_interval: Duration::from_secs(1),
+            no_perf_events: true,
+            perf_event_test_path: Path::new("").to_path_buf(),
```
TheElectronWill, line 95:
```suggestion
            perf_event_test_path: PathBuf::from(""),
```

TheElectronWill (changes requested):

plugin-rapl/Cargo.toml
```diff
@@ -17,6 +16,13 @@ log = "0.4.22"
 perf-event-open-sys2 = "5.0.6"
 regex = "1.10.6"
 serde = { version = "1.0.210", features = ["derive"] }
+once_cell = "1.21.3"
+
+[dev-dependencies]
+alumet = {path = "../alumet", features = ["test"]}
+toml = "0.8.19"
+tempfile = "3.20.0"
+nix = "0.30.1"
```
TheElectronWill, line 25: I think `nix` is no longer used :)

plugin-rapl/src/lib.rs
```diff
@@ -72,26 +99,27 @@ impl AlumetPlugin for RaplPlugin {
         }
 
         // Discover RAPL domains available in perf_events and powercap. Beware, this can fail!
-        let try_perf_events = perf_event::all_power_events();
-        let try_power_zones = powercap::all_power_zones();
+        let try_perf_events = self.get_all_power_events();
+        let try_power_zones = self.get_all_power_zones();
 
         let (available_domains, subset_indicator) = match (try_perf_events, try_power_zones) {
```
TheElectronWill, line 105: Can you move this into the `consistency` module? No need to refactor it extensively, just extract it outside of start, in a function that returns `available_domains` and `subset_indicator`.

plugin-rapl/Cargo.toml
ahuret: it's still used in perf_event.rs

plugin-rapl/src/perf_event.rs
```diff
@@ -38,57 +43,109 @@ pub struct PowerEvent {
...
+        }
+    }
+    Ok(events)
+}
+
+impl PowerEventFactory {
+    /// creates a new PowerEvent from an event base path. In case the path is not identified as a RAPL event one, None will be returned.
+    /// (eg: /sys/devices/power/events/energy-cores)
+    pub fn from_path(base_path: &Path) -> anyhow::Result<Option<PowerEvent>> {
+        if !Self::is_event_path(base_path) {
```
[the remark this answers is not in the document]
ahuret: not sure to understand this one..

before it was doing

```rust
        if path.is_file() && !file_name.contains('.') {
            // The files are named "energy-pkg", "energy-dram", ...
            if let Some(event_name) = file_name.strip_prefix("energy-") {
```

vs now

```rust
        path.is_file() && !file_name.contains('.') && file_name.strip_prefix("energy-").is_some()
```

In case first two conditions (`path.is_file() && !file_name.contains('.')` doesn't pass, rust will return before evaluating third one (`file_name.strip_prefix("energy-").is_some()`) in both case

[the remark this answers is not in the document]
ahuret: removed the is_event_path and now made the check in `name_from_base_path`

PR #198, opened 2025-06-02
fix(plugin-nvidia-nvml): change the way to collect process related measurements

ahuret:
Old way to collect process related measurements was buggy, cause it was using pid instead of a unix timestamp in `process_utilization_stats` method. This lead to unexpected values.

Now the source/probe will keep an Option<Timestamp> in memory that contains the timestamp of the last poll. This is helpful to ask `process_utilization_stats` the values since this timestamp.

Also added a new Unit called `Percent`.

### DIFF
--- a/plugin-nvidia-nvml/src/nvml/metrics.rs
+++ b/plugin-nvidia-nvml/src/nvml/metrics.rs
@@ -52,21 +52,10 @@ impl Metrics {
                 Unit::DegreeCelsius,
                 "Instantaneous temperature of the GPU at the time of the measurement",
             )?,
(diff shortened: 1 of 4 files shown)


plugin-nvidia-nvml/src/nvml/metrics.rs
```diff
@@ -53,21 +53,6 @@ impl Metrics {
                 "Instantaneous temperature of the GPU at the time of the measurement",
             )?,
             major_utilization_gpu: alumet.create_metric("nvml_gpu_utilization", Unit::Unity, "GPU rate utilization")?,
```
titouanj, line 55: Isn't 'rate' a percentage?

titouanj (dismissed):
I don't understand the change of metric registration order but ok

PR #200, opened 2025-06-04
feat(plugin-process-to-cgroup-bridge): create new plugin

ahuret:
Add a new **Transform** plugin called `process-to-cgroup-bridge`.

This plugin aims to be enabled when `plugin-nvidia-nvml` (input) is enable.

It will find the cgroup related to the process measurement and will set it as the measurement consumer.
The plugin come with configuration:

- ability to filter the metrics impacted by this plugin
- ability to merge the measurements coming from different processes but with same cgroup at same timestamp
- ability to not keep measurements that have been transformed

The plugin also comes with integration tests that have been placed under `src` because needed some `#[cfg(test)]` tags that are not available in `tests` subdir.

### DIFF
--- a/plugin-process-to-cgroup-bridge/Cargo.toml
+++ b/plugin-process-to-cgroup-bridge/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "plugin-process-to-cgroup-bridge"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+alumet = { path = "../alumet" }
+anyhow = "1.0.79"
+log = "*"
+serde = { version = "1.0.198", features = ["derive"] }
+
+[dev-dependencies]
+alumet = {path = "../alumet", features = ["test"]}
+toml = "0.8.19"
+tempfile = "3.20.0"
+lazy_static = "1.5.0"
--- a/plugin-process-to-cgroup-bridge/src/tests.rs
+++ b/plugin-process-to-cgroup-bridge/src/tests.rs
@@ -1,0 +1,451 @@
+// This file contains integration tests. Not having them in tests/ subdir is expected here.
+// This is to bypass a rust behavior while cfg(test) annotations are not visible for integration tests under tests/ subdir
+// In our context we want use these annotations to change how code is compiled under tests (most of time to mock constants)
+#[cfg(test)]
+use std::{
+    fs::{self, File},
+    io::Write,
+    path::PathBuf,
+    time::{Duration, UNIX_EPOCH},
+};
+
+use tempfile::{tempdir, TempDir};
+
+use crate::{Config, ProcessToCgroupBridgePlugin};
+
+use alumet::{
+    agent::{
+        self,
+        plugin::{PluginInfo, PluginSet},
+    },
+    measurement::{MeasurementAccumulator, MeasurementBuffer, MeasurementPoint, Timestamp, WrappedMeasurementValue},
+    metrics::TypedMetricId,
+    pipeline::naming::TransformName,
+    pipeline::{elements::error::PollError, elements::source::trigger::TriggerSpec, Source},
+    plugin::rust::AlumetPlugin,
+    plugin::PluginMetadata,
+    resources::{Resource, ResourceConsumer},
+    test::{runtime::TransformCheckInputContext, RuntimeExpectations},
+    units::Unit,
+};
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref T: Timestamp = Timestamp::from(UNIX_EPOCH);
+    static ref T2: Timestamp = Timestamp::from(UNIX_EPOCH + Duration::from_secs(1));
+}
+
+#[derive(Debug, Clone)]
+struct ExpectedCounts {
+    t_initial: usize,
+    t2_initial: usize,
+    t_shared: usize,
+    t_single: usize,
+    t2_shared: usize,
+    t2_single: usize,
+}
+
+fn count_measurements_by_consumer_and_time(
+    measurements: &MeasurementBuffer,
+    consumer_filter: impl Fn(&ResourceConsumer) -> bool,
+    timestamp: Timestamp,
+) -> usize {
+    measurements
+        .iter()
+        .filter(|p| consumer_filter(&p.consumer) && p.timestamp == timestamp)
+        .count()
+}
+
+fn count_initial_measurements(measurements: &MeasurementBuffer, timestamp: Timestamp) -> usize {
+    count_measurements_by_consumer_and_time(
+        measurements,
+        |consumer| {
+            matches!(
+                consumer,
+                ResourceConsumer::Process { .. } | ResourceConsumer::LocalMachine
+            )
+        },
+        timestamp,
+    )
+}
+
+fn count_cgroup_measurements(measurements: &MeasurementBuffer, cgroup_path: &str, timestamp: Timestamp) -> usize {
+    count_measurements_by_consumer_and_time(
+        measurements,
+        |consumer| matches!(consumer, ResourceConsumer::ControlGroup { path } if path == cgroup_path),
+        timestamp,
+    )
+}
+
+fn assert_measurement_counts(measurements: &MeasurementBuffer, expected: ExpectedCounts) {
+    let t_initial_count = count_initial_measurements(measurements, *T);
+    assert_eq!(t_initial_count, expected.t_initial);
+
+    let t2_initial_count = count_initial_measurements(measurements, *T2);
+    assert_eq!(t2_initial_count, expected.t2_initial);
+
+    let t_shared_count = count_cgroup_measurements(measurements, "/system.slice/shared.slice", *T);
+    assert_eq!(t_shared_count, expected.t_shared);
+
+    let t_single_count = count_cgroup_measurements(measurements, "/system.slice/single.slice", *T);
+    assert_eq!(t_single_count, expected.t_single);
+
+    let t2_shared_count = count_cgroup_measurements(measurements, "/system.slice/shared.slice", *T2);
+    assert_eq!(t2_shared_count, expected.t2_shared);
+
+    let t2_single_count = count_cgroup_measurements(measurements, "/system.slice/single.slice", *T2);
+    assert_eq!(t2_single_count, expected.t2_single);
+}
+
+fn run_test_with_config(config: Config, expected_counts: ExpectedCounts) -> anyhow::Result<()> {
+    let _ = prepare_procfs_mock()?;
+    let mut plugins = PluginSet::new();
+
+    plugins.add_plugin(PluginInfo {
+        metadata: PluginMetadata::from_static::<DumbNvmlPlugin>(),
+        enabled: true,
+        config: None,
+    });
+    plugins.add_plugin(PluginInfo {
+        metadata: PluginMetadata::from_static::<ProcessToCgroupBridgePlugin>(),
+        enabled: true,
+        config: Some(config_to_toml_table(&config)),
+    });
+
+    let make_input = move |ctx: &mut TransformCheckInputContext| -> MeasurementBuffer {
+        prepare_mock_measurements(ctx).expect("failed to prepare mock points")
+    };
+
+    let check_output = move |measurements: &MeasurementBuffer| {
+        assert_measurement_counts(measurements, expected_counts.clone());
+    };
+
+    let runtime_expectations = RuntimeExpectations::new().test_transform(
+        TransformName::from_str("process-to-cgroup-bridge", "transform"),
+        make_input,
+        check_output,
+    );
+
+    let agent = agent::Builder::new(plugins)
+        .with_expectations(runtime_expectations)
+        .build_and_start()
+        .unwrap();
+
+    agent.wait_for_shutdown(Duration::from_secs(2)).unwrap();
+
+    Ok(())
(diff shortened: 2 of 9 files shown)


plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
+#[cfg(test)]
```
TheElectronWill, line 1: Add a small comment explaining why this cannot be an integration test

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+mod tests {
```
TheElectronWill, line 2: You don't need the `mod tests` here, it's already a module :)
In Rust, each file is a module.

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+use std::sync::OnceLock;
+
+#[cfg(test)]
+static PROC_PATH: OnceLock<PathBuf> = OnceLock::new();
+
+#[cfg(test)]
+fn get_proc_path() -> &'static PathBuf {
+    PROC_PATH.get_or_init(|| {
+        panic!("PROC_PATH not initialized in test - call set_proc_path_for_test() first");
+    })
```
TheElectronWill, line 36:
```suggestion
    PROC_PATH.get().expect("PROC_PATH must be initialized at the beginning of the test: call set_proc_path_for_test()")
```

TheElectronWill: if you don't init, use get + expect

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+            t2_single: 0,
+        };
+
+        run_test_with_config(config, expected)
+    }
+
+    #[test]
+    fn test_process_id_not_found_in_procfs() -> anyhow::Result<()> {
+        let base_path = prepare_procfs_mock()?;
+        set_proc_path_for_test(base_path);
```
TheElectronWill, line 222: Tests run in parallel, and `set_proc_path_for_test` will only initialize the global variable once (the first test wins). Therefore, the `base_path` is only correct for the test that wins the race, which is non deterministic.
If you intend the first test to win, `prepare_procfs_mock` should be called only once, on initialization. I would replace the two lines with something like
```rs
let base_path = prepare_procfs_for_test()
```

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        if !Path::new(&procfs_cgroup_filepath).exists() {
+            return Ok(None);
+        }
+
+        let contents = fs::read_to_string(&procfs_cgroup_filepath)
+            .with_context(|| format!("Failed to read {:?}", procfs_cgroup_filepath))?;
+
+        // a typical procfs cgroup file will contain only one line
+        // eg: 0::/system.slice/docker-7c7fc86f5f2a609c41c6edd65bd1b64135124a687fa6516f6b177b040d6e3b68.scope
```
TheElectronWill, line 138: This is only true for cgroup v2. For cgroup v1, it contains one line per cgroup hierarchy.
Example:
```
2:memory:/daemons
5:cpuacct,cpu:/daemons
```

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+        let procfs_cgroup_base_path = get_proc_path();
+
+        let procfs_cgroup_filepath = procfs_cgroup_base_path.join(pid.to_string()).join("cgroup");
+
+        if !Path::new(&procfs_cgroup_filepath).exists() {
+            return Ok(None);
+        }
+
+        let contents = fs::read_to_string(&procfs_cgroup_filepath)
+            .with_context(|| format!("Failed to read {:?}", procfs_cgroup_filepath))?;
```
TheElectronWill, line 135: error messages should start with a lowercase letter

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn find_cgroup_path_from_process_id(pid: u32) -> anyhow::Result<Option<String>> {
+        let procfs_cgroup_base_path = get_proc_path();
+
+        let procfs_cgroup_filepath = procfs_cgroup_base_path.join(pid.to_string()).join("cgroup");
+
+        if !Path::new(&procfs_cgroup_filepath).exists() {
+            return Ok(None);
+        }
+
+        let contents = fs::read_to_string(&procfs_cgroup_filepath)
```
TheElectronWill, line 134: if the file doesn't exist, it would be more logical to return an error

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+            let parts: Vec<&str> = line.split(':').collect();
+            if parts.len() >= 3 {
+                let cgroup_path = parts[2];
+                if !cgroup_path.is_empty() {
+                    return Ok(Some(cgroup_path.to_string()));
+                }
+            }
+        }
+
+        Ok(None)
```
TheElectronWill, line 149: if the file's format is invalid, it would be more logical to return an error

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    fn cgroup_measurement_from_measurement(
+        &self,
+        measurement: &MeasurementPoint,
+    ) -> anyhow::Result<Option<MeasurementPoint>> {
+        if !self.processes_metrics_ids.contains(&measurement.metric) {
+            return Ok(None);
+        }
+        let pid = match Self::extract_process_id_from_measurement(measurement) {
```
TheElectronWill, line 105: If you ignore the error of `extract_process_id_from_measurement`, it should probably return `Option` instead.

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,199 @@
...
+
+            let mut new_point = group[0].clone(); // cannot panic since group cannot be empty
+            new_point.value = WrappedMeasurementValue::F64(mean);
+            aggregates.push(new_point);
+        }
+        aggregates
+    }
+}
+
+fn extract_numeric_value(measurement: &MeasurementPoint) -> Option<f64> {
```
TheElectronWill, line 194: The other "utilities" functions could be outside of the impl too, this way you would not need `Self::` to call them.

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+                    let mut file = File::create(full_path)?;
+                    file.write_all(content.as_bytes())?;
+                }
+            }
+        }
+        Ok(())
+    }
+
+    // Mock plugin implementations
+    struct DumbNvmlPlugin;
```
TheElectronWill, line 390: nice

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        Ok(m)
+    }
+
+    fn config_to_toml_table(config: &Config) -> toml::Table {
+        toml::Value::try_from(config).unwrap().as_table().unwrap().clone()
+    }
+
+    // Mock filesystem utilities
+    enum EntryType<'a> {
```
TheElectronWill, line 362: You could use `&'static str` in these strucs to avoid the lifetime parameter. It would work since you are using constant strings in your test. But it's fine the way it is too.

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+            .unwrap();
+
+        agent.wait_for_shutdown(Duration::from_secs(2)).unwrap();
+
+        Ok(())
+    }
+
+    fn prepare_procfs_mock() -> anyhow::Result<PathBuf> {
+        let tmp = tempdir()?;
+        let base_path = tmp.keep();
```
TheElectronWill, line 286: The tempdir should be returned from the function, so that it is cleared after the test. Otherwise it will clutter /tmp.

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+    #[test]
+    fn test_process_id_not_found_in_procfs() -> anyhow::Result<()> {
+        let base_path = prepare_procfs_mock()?;
+        set_proc_path_for_test(base_path);
+
+        let mut plugins = PluginSet::new();
+
+        let source_config = Config {
+            processes_metrics: vec!["metric_a".to_string(), "metric_b".to_string()],
+            ..Default::default()
```
TheElectronWill, line 228: beautiful

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        agent.wait_for_shutdown(Duration::from_secs(2)).unwrap();
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_default_setup() -> anyhow::Result<()> {
+        let config = Config {
+            processes_metrics: vec!["metric_a".to_string(), "metric_b".to_string()],
```
TheElectronWill, line 144: The following util function could make it nicer to read:
```rs
fn string_vec(values: &[&str]) -> Vec<String> {
  values.into_iter().map(|s| s.to_string()).collect()
}

// example
processes_metrics: string_vec(&["metric_a", "metric_b"])
```

plugin-process-to-cgroup-bridge/src/transform.rs
TheElectronWill: Also, read_to_string already check if it exists. You can detect that by checking the `ErrorKind`. See https://doc.rust-lang.org/std/io/struct.Error.html

plugin-process-to-cgroup-bridge/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "plugin-process-to-cgroup-bridge"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+alumet = { path = "../alumet" }
+anyhow = "1.0.79"
+log = "*"
+serde = { version = "1.0.198", features = ["derive"] }
+serde_json = "1.0"
```
TheElectronWill, line 11: Do you need serde_json?

plugin-process-to-cgroup-bridge/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        // metric_b points
+        m.push(create_point(*T, metric_b, ResourceConsumer::Process { pid: 1 }, 10));
+        m.push(create_point(*T, metric_b, ResourceConsumer::Process { pid: 2 }, 10));
+
+        Ok(m)
+    }
+
+    fn config_to_toml_table(config: &Config) -> toml::Table {
+        toml::Value::try_from(config).unwrap().as_table().unwrap().clone()
```
TheElectronWill, line 358: IIRC you can use `toml::Table::try_from` directly

TheElectronWill (changes requested):

plugin-process-to-cgroup-bridge/src/tests.rs
ahuret: Actually both function have their own responsability.

In the context of extract_process_id_from_measurement, failing/returning error when the measurement has not the right format sounds reasonable.

In the context of cgroup_measurement_from_measurement which call the precedent function, it's also reasonable to "mute" the error because it's expected to have measurements that are not "process" ones and we decide to not treat them (without failing).

plugin-process-to-cgroup-bridge/src/tests.rs
ahuret: Removed the global variable that was shared by all tests. Instead relying on configuration only available for tests.
We had multiple issues by using a global variable as tests are run in //. We were not able to make multiple mocks for multiple tests. And also it was complicated to have TmpDir deleted as the Dir was shared by all.

plugin-process-to-cgroup-bridge/src/transform.rs
```diff
@@ -0,0 +1,187 @@
...
+        // eg:
+        //     2:memory:/daemons
+        //     5:cpuacct,cpu:/daemons
+        // in this case we take the first one arbitrary and log a warning
+        // todo: implement a matrix between metric names and cgroup v1 controllers to be able to select the right cgroup line
+        for line in contents.lines() {
+            let parts: Vec<&str> = line.split(':').collect();
+            if parts.len() >= 3 {
+                if parts[1] != "" {
+                    log::warn!("detected that pid {pid} is cgroup v1 based, selecting arbitrary the cgroup");
```
TheElectronWill, line 116:
```suggestion
                    log::warn!("detected that process {pid} is managed by cgroup v1, selecting the cgroup arbitrarily");
```
I think you meant "arbitrarily" ^^

TheElectronWill (approved):

PR #202, opened 2025-06-10
feat(plugin-grace-hopper): Energy instead of power

AngeCyp:
- Add the energy instead of power in the return of plugin and also improve code test coverage
- Add some little functions in the `Timestamp` wrapper for easier use.

Tarpaulin output:
image

### DIFF
--- a/plugin-grace-hopper/src/lib.rs
+++ b/plugin-grace-hopper/src/lib.rs
@@ -118,7 +118,7 @@ fn get_sensor_from_dir(entry: DirEntry) -> Result<Option<Sensor>, anyhow::Error>
         return Ok(None);
     }
     let file = File::open(&device_file).context("failed to open the file")?;
-    let (kind, socket) = get_sensor_information_from_file(file)?;
+    let (kind, socket) = get_sensor_information_from_file(&file)?;
     Ok(Some(Sensor {
         kind,
         socket,
@@ -135,8 +135,8 @@ fn get_sensor_from_dir(entry: DirEntry) -> Result<Option<Sensor>, anyhow::Error>
 /// Returns a [Result] containing a tuple `(String, u32)` on success
 /// - The first element of the tuple (`String`) represents the sensor type (e.g., "grace", "cpu").
 /// - The second element (`u32`) represents the associated socket number.
-fn get_sensor_information_from_file(file: File) -> Result<(String, u32), anyhow::Error> {
-    let reader = io::BufReader::new(&file);
+fn get_sensor_information_from_file(file: &File) -> Result<(String, u32), anyhow::Error> {
+    let reader = io::BufReader::new(file);
     let mut lines = reader.lines();
     let kind: String;
     let socket: u32;
@@ -187,8 +187,8 @@ mod tests {
     use super::*;
     use crate::GraceHopperPlugin;
     use anyhow::Result;
-    use std::fs::File;
-    use std::io::Write;
+    use std::fs::{File, OpenOptions};
+    use std::io::{Seek, Write};
     use std::time::Duration;
     use tempfile::tempdir;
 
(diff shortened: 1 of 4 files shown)


TheElectronWill (changes requested):

plugin-grace-hopper/src/probe.rs
```diff
@@ -62,16 +63,18 @@ impl Source for GraceHopperProbe {
...
                 self.metric
                     .expect("can't push to the MeasurementAccumulator because can't retrieve the metric"),
                 Resource::CpuPackage { id: self.socket },
                 self.consumer.clone(),
-                power,
+                computed_energy,
             )
-            .with_attr("sensor", self.kind.clone()),
+            .with_attr("sensor", self.kind.clone())
+            .with_attr("min_interval_update_ms", self.power_stats_interval.as_millis() as f64),
```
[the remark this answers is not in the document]
AngeCyp: This is the minimum interval for a power telemetry update. I mean, if the value is 50, it means that we have to wait at least 50 ms before updating the power value. See [here](https://docs.nvidia.com/grace-perf-tuning-guide/power-thermals.html#grace-power-telemetry)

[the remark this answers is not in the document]
AngeCyp: Not added anymore, just printed through `log::info!()` and stored.

plugin-grace-hopper/src/probe.rs
```diff
@@ -100,14 +102,45 @@ pub fn read_power_value(buffer: &mut String, file: &mut File) -> Result<u64, any
...
+    let time_elapsed_opt = match last_timestamp.update(final_value) {
+        CounterDiffUpdate::FirstTime => None,
+        CounterDiffUpdate::Difference(diff) | CounterDiffUpdate::CorrectedDifference(diff) => Some(diff),
+    };
+    if let Some(time_elapsed_ns) = time_elapsed_opt {
+        let time_in_seconds_u64 = time_elapsed_ns / 1_000_000_000;
+        let time_in_seconds = time_in_seconds_u64 as f64;
+        Ok(power as f64 * time_in_seconds)
+    } else {
+        Ok(power as f64)
```
ahuret, line 129: in case counterdiff is None (here it's the case) I think returning power as is since is wrong because we still want an energy, so I would rather return an Ok Option None

ahuret (changes requested):

plugin-grace-hopper/src/probe.rs
```diff
@@ -61,19 +80,29 @@ impl GraceHopperProbe {
...
                 timestamp,
                 self.metric
                     .expect("can't push to the MeasurementAccumulator because can't retrieve the metric"),
                 Resource::CpuPackage { id: self.socket },
                 self.consumer.clone(),
-                power,
+                computed_energy,
             )
             .with_attr("sensor", self.kind.clone()),
         );
```
ahuret, line 105:
```rust
        if self.last_measure.is_some() {
            let computed_energy = self.last_measure.as_mut().unwrap().compute_energy(&new_measure)?;
            measurements.push(
                MeasurementPoint::new(
                    timestamp,
                    self.metric
                        .expect("can't push to the MeasurementAccumulator because can't retrieve the metric"),
                    Resource::CpuPackage { id: self.socket },
                    self.consumer.clone(),
                    computed_energy,
                )
                .with_attr("sensor", self.kind.clone()),
            );
        }
        self.last_measure = Some(PowerMeasure {
            timestamp: new_measure.timestamp,
            power: new_measure.power,
        });
```

ahuret: this way you define only one time the last_measure values, wdyt ?

ahuret (approved):

plugin-grace-hopper/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub struct Sensor {
     /// Socket associated to the sensor
     socket: u32,
     /// How often value are updated
-    average_interval: Duration,
+    _average_interval: Duration,
```
titouanj, line 32: Comment TODO:
```suggestion
    // TODO: average_interval: Duration,
```

PR #204, opened 2025-06-12
fix: added units to the prometheus and opentelemetry exporters

guigomcha:
By adding the units also to the gauge we get grafana to automatically add the axis label with the unit

image

### DIFF
--- a/plugin-opentelemetry/src/output.rs
+++ b/plugin-opentelemetry/src/output.rs
@@ -1,5 +1,6 @@
 use alumet::{
     measurement::{MeasurementBuffer, WrappedMeasurementValue},
+    metrics::Metric,
     pipeline::elements::{error::WriteError, output::OutputContext},
 };
 use anyhow::Context;
@@ -97,16 +98,8 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
                 "{}{}{}",
                 self.prefix,
                 if self.append_unit_to_metric_name {
-                    let unit_string = if self.use_unit_display_name {
-                        full_metric.unit.display_name()
-                    } else {
-                        full_metric.unit.unique_name()
-                    };
-                    if unit_string.is_empty() {
-                        full_metric.name.to_owned()
-                    } else {
-                        format!("{}_{}", full_metric.name, unit_string)
-                    }
+                    let unit_string = get_unit_string(full_metric, self.use_unit_display_name);
+                    format!("{}_{}", full_metric.name, unit_string)
                 } else {
                     full_metric.name.clone()
                 },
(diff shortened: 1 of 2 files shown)


plugin-prometheus-exporter/src/output.rs
```diff
@@ -152,3 +151,15 @@ fn sanitize_name(name: String) -> String {
...
+
+fn get_unit_string(full_metric: &Metric, use_unit_display_name: bool) -> String {
+    let unit_string = if use_unit_display_name {
+        full_metric.unit.display_name()
+    } else {
+        full_metric.unit.unique_name()
+    };
+    if unit_string.is_empty() {
+        return full_metric.name.to_owned();
+    }
```
ahuret, line 163: same here

plugin-opentelemetry/src/output.rs
```diff
@@ -97,16 +98,8 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
                 "{}{}{}",
```
ahuret, line 98: As I see in your screenshot, this will duplicate the unit at two different place.
Say the original metric name is `mem_total` and the unit is `kB` if I understand right with your screenshot, the unit name will be duplicated both using your own logic and using prometheus/opentelemetry logic.

`mem_total_kB_alumet_kB` where first `kB` is set by you and second `kB` is set by prometheus/opentelemetry.
I believe that it make more sense to have only one and to follow prometheus/openetelemetry naming convention this would be the one at the end .

guigomcha: yep, now that I add the unit to the prom/otpl object I don't need to manually add it to the name. I'll delete it

plugin-opentelemetry/src/output.rs
```diff
@@ -154,3 +147,15 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
+
+fn get_unit_string(full_metric: &Metric, use_unit_display_name: bool) -> String {
+    let unit_string = if use_unit_display_name {
+        full_metric.unit.display_name()
+    } else {
+        full_metric.unit.unique_name()
+    };
+    if unit_string.is_empty() {
+        return full_metric.name.to_owned();
+    }
```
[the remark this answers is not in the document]
guigomcha: it comes from https://github.com/alumet-dev/alumet/blob/main/plugin-csv/src/output.rs#L105 tbh, maybe @TheElectronWill can explain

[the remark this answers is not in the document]
TheElectronWill: The full context is:
```
let metric_name = if self.append_unit_to_metric_name {
                let unit_string = if self.use_unit_display_name {
                    full_metric.unit.display_name()
                } else {
                    full_metric.unit.unique_name()
                };
                if unit_string.is_empty() {
                    full_metric.name.to_owned()
                } else {
                    format!("{}_{}", full_metric.name, unit_string)
                }
            } else {
                full_metric.name.clone()
            };
```
In the original code, it was just an optimization to avoid `format!` when there is no unit.

Here, `get_unit_string` should not do this: it should only return the unit's name.

TheElectronWill (changes requested):

Issue #216, opened 2025-07-02
Emit an event before the pipeline stops / when the `exec` child stops


TheElectronWill:
@millillitre needs to perform a measurement at the end of the experiment.
`Plugin::stop` is not suited for this, as it is called _after_ the pipeline elements (sources, transforms, outputs) are stopped and dropped.

Potential solutions:
- Add `Plugin::pre_stop`.
- Add a dedicated EventBus.

For now, I'd prefer the second solution.

TheElectronWill:
Hi! Yes, I was thinking about something like that :)

TheElectronWill:
Solved by #222

PR #222, opened 2025-07-15
feat(core): add event to trigger measurement at experiment end

millillitre:
Adding a dedicated EventBus to perform a measurement at the end of the experiment because Plugin::stop is not suited for this, as it is called after the pipeline elements (sources, transforms, outputs) are stopped and dropped.

### DIFF
--- a/alumet/src/plugin/event.rs
+++ b/alumet/src/plugin/event.rs
@@ -165,7 +165,7 @@ pub struct StartResourceMeasurement(pub Vec<Resource>);
 
 /// Event occurring when measurements should be performed at the end of the consumer experiment.
 #[derive(Clone)]
-pub struct EndConsumerMeasurement();
+pub struct EndConsumerMeasurement;
 
 impl Event for StartConsumerMeasurement {}
 impl Event for StartResourceMeasurement {}
--- a/alumet/src/agent/exec.rs
+++ b/alumet/src/agent/exec.rs
@@ -58,7 +58,7 @@ pub fn watch_process(
 
     // Publish an event to perform a measurement at the end of the experiment
     log::info!("Publishing EndConsumerMeasurement event");
-    crate::plugin::event::end_consumer_measurement().publish(EndConsumerMeasurement());
+    crate::plugin::event::end_consumer_measurement().publish(EndConsumerMeasurement);
 
     // Stop the pipeline
     agent.pipeline.control_handle().shutdown();


millillitre:
> If I understand well, it's related to issue #216 am I correct ?

Yes this is right, it's related to this issue.

AngeCyp (commented):
As I see, you only created the dedicated bus event, but you didn't implement the "running part". Maybe you could try to do it ? 
I mean:  `start_consumer_measurement` is used [there](https://github.com/alumet-dev/alumet/blob/fb98590f01be4b49673554324105cb63f04b144f/alumet/src/agent/exec.rs#L74) 
But your new `end_consumer_measurement()` function is not used.
I don't know what was planned on that side. Maybe you don't have to do it? Tell me if I'm wrong

millillitre:
> As I see, you only created the dedicated bus event, but you didn't implement the "running part". Maybe you could try to do it ? I mean: `start_consumer_measurement` is used [there](https://github.com/alumet-dev/alumet/blob/fb98590f01be4b49673554324105cb63f04b144f/alumet/src/agent/exec.rs#L74) But your new `end_consumer_measurement()` function is not used. I don't know what was planned on that side. Maybe you don't have to do it? Tell me if I'm wrong

Thanks for your review, I see I missed some information! You're right about the `end_consumer_measurement` function. Initially, I focused on the event bus setup and didn't extend the implementation to other parts of the codebase, such as `src/agent`, as I was primarily working within `src/plugin` repository. The plan was to be able to capture the end of the process before it exits, so I will try to integrate the `end_consumer_measurement function` [here](https://github.com/alumet-dev/alumet/blob/fb98590f01be4b49673554324105cb63f04b144f/alumet/src/agent/exec.rs#L74) as well this week.

millillitre:
I add an event publication mechanism to signal the end of a consumer measurement process by adding the capability to publish an `EndConsumerMeasurement` event just before shutting down the pipeline. I hope it is what you told me to do.

AngeCyp:
Hello, sorry for being late, This looks good to me, but I will not approve before @TheElectronWill review just to be sure. 
Furthermore, did you do: 

- `cargo fmt --check`
- `cargo clippy --no-deps`

For linting and improving your code ? (Don't worry about the previously written code which not satisfying clippy) 

-----------

I'll also have to ask you to to modify your commits according to [conventional commit rules](https://www.conventionalcommits.org/en/v1.0.0/).
For example, your scope (the scope is optional) could be "core":

> feat(core): updating endevent implementation without too much logging

Here is the regex that check if it's compliant:
> (^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)(\([a-zA-Z0-9_\-\.]+\))?(\!)?: ([a-z0-9_ \.\-\`])+((\n|\r\n){2}[\s\S]+)?)

AngeCyp:
Thanks for running both tools on your code.
You modified two commits, they should pass the CI check, good job, but there is still three commits:
- c2e81ae70150e89293d1938b10ba8efe5b72e9b4
- a32502646945bc41b14758ec3c59164f27a59fb8
- c4437c3abfc38342d91cb06aef8d9b751b055b1b

which are not compliant, you should modify them too, you could also squash them, it's as you want.

readme/OUTPUT_PLUGIN_README_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+# Plugin's name
```
TheElectronWill, line 1: I think the `readme` folder has been added by mistake ^^

alumet/src/plugin/event.rs
```diff
@@ -155,8 +163,13 @@ pub struct StartConsumerMeasurement(pub Vec<ResourceConsumer>);
 #[derive(Clone)]
 pub struct StartResourceMeasurement(pub Vec<Resource>);
 
+/// Event occurring when measurements should be performed at the end of the consumer experiment.
+#[derive(Clone)]
+pub struct EndConsumerMeasurement();
```
TheElectronWill, line 168: Since this struct carries no data, you can declare it like this:
```rs
pub struct EndConsumerMeasurement;
```
and use it like this: `EndConsumerMeasurement` (without the parentheses).

TheElectronWill (changes requested):

TheElectronWill (approved):

PR #228, opened 2025-07-29
feat(plugin-cgroup): "Raw" cgroup plugin

### LINKED ISSUE #83: Support "raw" cgroups
We currently support K8S (cgroups v2), OAR2 (cgroups v1) and OAR3 (cgroups v2), with a roadmap for Slurm.
But there is no plugin for "raw" cgroups, i.e. Linux cgroups without K8S nor OAR information.

TheElectronWill:
Based on `feature/new-cgroups`.

Resolves #83.

### DIFF
--- a/util-cgroups/src/file_watch/inotify.rs
+++ b/util-cgroups/src/file_watch/inotify.rs
@@ -16,7 +16,7 @@ use nix::{
     sys::inotify::{AddWatchFlags, InitFlags, Inotify, InotifyEvent, WatchDescriptor},
 };
 use rustc_hash::{FxBuildHasher, FxHashMap, FxHashSet};
-use walkdir::WalkDir;
+use walkdir::{DirEntry, WalkDir};
 
 pub struct InotifyWatcher {
     thread_handle: Option<JoinHandle<()>>,
@@ -41,7 +41,7 @@ impl InotifyWatcher {
         let mut deleted = Vec::new();
         for path in paths_to_watch {
             match path.metadata().map(|m| m.is_dir()) {
-                Ok(true) => watch.watch_recursively(path, &mut created, &mut deleted)?,
+                Ok(true) => watch.watch_recursively(path, &mut created, &mut deleted),
                 Ok(false) => return Err(anyhow::anyhow!("cannot watch {path:?} because it is not a directory")),
                 Err(e) => return Err(e).context(format!("cannot watch {path:?}")),
             }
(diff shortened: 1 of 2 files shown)


util-cgroups/src/file_watch/inotify.rs
```diff
@@ -177,7 +177,7 @@ impl<E: EventHandler> WatchLoop<E> {
                         self.unmark_watched(&path);
                     }
                     Err(e) => {
-                        return Err(anyhow::Error::new(e).context(format!("could not add watch to path {path:?}")));
+                        log::error!("could not add watch to path {path:?}: {e:#}");
```
AngeCyp, line 180: The function seems to always return ok, why don't we continue to take advantage of `Err` from `Result` ?

AngeCyp: If no error is return, why should'nt replace `?` with `unwrap` ? https://github.com/alumet-dev/alumet/blob/b165e07ea16270adf76520a2720536058d9bb2e0/util-cgroups/src/file_watch/inotify.rs#L44

util-cgroups/src/hierarchy.rs
```diff
@@ -388,6 +389,25 @@ pub enum HierarchyError {
...
+    // check if the the `app.slice` cgroup, which contains cgroups for user applications, exists
+    let path = v2_hierarchy_root.join(format!("user.slice/user-{euid}.slice/user@{euid}.service/app.slice"));
+    match std::fs::exists(&path) {
+        Ok(true) => Ok(path),
+        Ok(false) => Err(anyhow::Error::new(std::io::Error::from(ErrorKind::NotFound))
+            .context(format!("not found in cgroupfs: {path:?}"))),
+        Err(e) => Err(anyhow::Error::new(e).context(format!("failed to check the existence of {path:?}"))),
+    }
+}
+
```
AngeCyp, line 410: Maybe add a little test for this function to show it works could be a good thing ?

util-cgroups/src/measure/v2/cpu.rs
```diff
@@ -23,19 +23,19 @@ struct CpuStatMapping {
 
 #[derive(Debug, Serialize)]
 pub struct CpuStatCollectorSettings {
-    pub usage: bool,
-    pub user: bool,
-    pub system: bool,
+    pub usage_usec: bool,
+    pub user_usec: bool,
+    pub system_usec: bool,
```
AngeCyp, line 28: Should be also corrected in other branches

plugin-cgroup/tests/raw.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // stop the pipeline and wait for it to terminate
+    handle.shutdown();
+    agent.wait_for_shutdown(TIMEOUT)?;
+    Ok(())
+}
+
+fn config_to_toml_table(config: &Config) -> toml::Table {
+    toml::Value::try_from(config).unwrap().as_table().unwrap().to_owned()
+}
```
AngeCyp, line 78: This function is used everywhere and in near all tests, maybe we could write it once somewhere and import it each time to use it?

util-cgroups/src/hierarchy.rs
TheElectronWill: it's used in integration tests somewhere

util-cgroups/src/file_watch/inotify.rs
TheElectronWill: The error may be improved later, e.g. if the root directory is inaccessible.

TheElectronWill: done.

PR #245, opened 2025-08-12
docs(plugin-procfs): README modify

victoryeagle77:


titouanj (changes requested):
Add link to the man pages about the procfs files/fields that are used

plugin-procfs/README.md
```diff
@@ -1,3 +1,112 @@
 # Procfs plugin
 
-Collects processes- and system-related metrics by reading the `/proc` virtual filesystem.
+Collects processes and system-related metrics by reading the [proc](<https://www.kernel.org/doc/html/latest/filesystems/proc.html>) virtual filesystem on UNIX based operating systems.
+
+## Requirements
+
+- UNIX operating system.
+- All metrics of the plugin needs to access to the process file system `/proc/sys/` dynamically generated by the kernel.
```
TheElectronWill, line 8: What is `/proc/sys`? We aren't reading from that.

What the plugin needs is read access to `/proc`. Depending on the situation, it will monitor more or less processes. For instance, running the Alumet agent as root will allow the procfs plugin to monitor every process of every user. Running the Alumet agent as user `gerard` will make it monitor the processes of this user. The permission to read `/proc/pid` for other processes can also be granted with capabilities. See https://docs.kernel.org/filesystems/proc.html#process-specific-subdirectories

For this PR, it would be great to make a separate section about permissions and capabilities, and put only one sentence in the list item.

TheElectronWill (changes requested):

plugin-procfs/README.md
TheElectronWill: To be more exhaustive, `/proc` access primarily depends on the `hidepid` mount option: https://docs.kernel.org/filesystems/proc.html#mount-options

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
 
-Collects processes- and system-related metrics by reading the `/proc` virtual filesystem.
+Collects processes and system-related metrics by reading the [proc](<https://www.kernel.org/doc/html/latest/filesystems/proc.html>) virtual filesystem on Linux based operating systems.
+
+## Requirements
+
+- Linux operating system.
+- Read access to file system `/proc/`.
+- Read access to file system `/proc/<pid>` for process stats.
+- Access mode in [/proc/mounts](<https://docs.kernel.org/filesystems/proc.html#mount-options>) file by configuring the [hidepid](#hidepid) parameter.
```
TheElectronWill, line 10: All three are the same item.
```suggestion
- Read access to the `/proc` virtual file system. Depending on the mount options, [some privileges might be needed](#procfs-access).
```

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+
+### hidepid
```
TheElectronWill, line 12:
```suggestion
### Procfs Access
```

And then explain what this parameter is and what it does, and add a link to the kernel docs.
Also, explain how to read `/proc/<pid>` when the parameter is high. As a user, do I need to be root? Do I need to run setcap? Something else?
Also, you can move this section to the bottom of the file.

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+|2|`Invisible` : All `/proc/<pid>/` will be fully invisible to other users|
+|4|`PTraceable` : Procfs should only contain `/proc/<pid>/` directories that the caller can ptrace|
+
+## Metrics
+
+There are various information collected by this plugin relative to Kernel, CPU, memory and processes :
+
+|Name|Type|Unit|Description|Ressource|RessourceConsumer|Attributes|
+|----|----|----|-----------|---------|-----------------|----------|
+|`kernel_cpu_time`|CounterDiff|MilliSecond|Time during the CPU is busy|LocalMachine|LocalMachine|[cpu_state](#cpu_state)|
```
TheElectronWill, line 29: Units should be written with their standard name, for example `ms` or `millisecond` for milliseconds, not `MilliSecond`.

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+
+#### Kind
+
+The kind of the memory is the allocated memory space reserved by the system or the hardware :
+
+|Value|Description|
+|-----|-----------|
+|`resident`|Physical memory actually used by a process|
+|`shared`|Part of memory used simultaneously by several processes|
+|`virtual`|Total virtual memory address space requested by a process|
```
TheElectronWill, line 60: The descriptions here are a bit simplistic, it's more complicated than that. IMO, it would be better to stay factual and use the same description as the manual. See the comments in the `procfs-core` crate, structure `StatM`, and [`man proc/pid/statm`](https://man7.org/linux/man-pages/man5/proc_pid_statm.5.html) and [`man /proc/pid/status`](https://man7.org/linux/man-pages/man5/proc_pid_status.5.html).

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+
+The kind of the CPU time delta is the average CPU time spent by various tasks :
+
+|Value|Description|
+|-----|-----------|
+|`user`|CPU time used to run userspace processes (not Kernel system processes)|
```
TheElectronWill, line 66: Same remark here: the descriptions are not accruate. Read the doc of the `procfs-core`, structure `Stat` (in `stat.rs`). The "user" kind is not the "time used to run userspace processes", but the time _spent_ by the process in _user mode_. Every process can have some `user` time and some `system` time.

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+|`nice`|CPU time consumed by tasks that give way to other higher priority processes|
+|`system`|CPU time used by the system (Kernel)|
+|`idle`|Time the CPU is in sleep mode or waiting for tasks|
+|`irq`|CPU time spent handling hardware interrupts (input/output or peripheral signals)|
+|`softirq`|CPU time spent handling software interrupts (delayed interrupts scheduled by the kernel)|
+|`steal`|Time a hypervisor uses CPU for other virtual machines on the same host|
```
TheElectronWill, line 82: some descriptions are in present tense, some are in past tense, please make everything consistent ^^ Use the past everywhere, so `used`, not `uses`.

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+
+## Configuration
+
+Here is a configuration example of the plugin. It's part of the ALUMET configuration file (eg: `alumet-config.toml`).
+
+```rust
+[plugins.procfs]
+poll_interval = "1s"
+flush_interval = "5s"
+```
```
TheElectronWill, line 94: This is still incorrect. Have you generated the config with `alumet-agent`?

TheElectronWill: You need to explain that, by default, `enabled` is set to true for each sub-section.

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+[plugins.procfs.processes]
+enabled = true
+refresh_interval = "2s"
+strategy = "watcher"
+```
+
+Likewise, you can collect processes groups metrics :
+
+```rust
+[[plugins.procfs.processes.groups]]
```
TheElectronWill, line 134: What does it do?

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
...
+    "MemAvailable",
+    "Cached",
+    "SwapCached",
+    "Active",
+    "Inactive",
+    "Mapped",
+]
+```
+
+You can active collect of processes metrics for the plugin with 2 different strategies (system watcher or internal event) :
```
TheElectronWill, line 122: And what does it change?
Imagine you're a new user, that has never used Alumet nor the plugin. You would ask yourself: What are the differences between the strategies? Why do they exist? What should I choose?

TheElectronWill (changes requested):

plugin-procfs/README.md
```diff
@@ -1,3 +1,138 @@
 # Procfs plugin
 
-Collects processes- and system-related metrics by reading the `/proc` virtual filesystem.
+Collects processes and system-related metrics by reading the [proc](<https://www.kernel.org/doc/html/latest/filesystems/proc.html>) virtual filesystem on Linux based operating systems.
+
+## Requirements
+
+- Linux operating system.
```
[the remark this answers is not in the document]
victoryeagle77: Yes : "https://www.cyberciti.biz/faq/howto-freebsd-mount-procfs/"

plugin-procfs/README.md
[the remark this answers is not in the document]
victoryeagle77: I don't how you can see that, I have already fixed this part yesterday

plugin-procfs/README.md
```diff
@@ -1,3 +1,139 @@
...
+|`guest`|Ticks spent running a virtual CPU for guest operating systems under control of the linux kernel|
+|`guest_nice`|Ticks spent running a niced guest|
+
+## Configuration
+
+Here is a configuration example of the plugin. It's part of the ALUMET configuration file (eg: `alumet-config.toml`). This plugin is composed of sections and each of them must be enabled or disabled with the `enabled` parameter to be set to true or false.
+
+To active the plugin to collect metrics relative to kernel utilization :
+
+```rust
```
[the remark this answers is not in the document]
victoryeagle77: Okay I will fix that but your comments here "Still missing comments/docs. You don't have to create them, they're already in the code" is for an older commit I already fixe that

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+
+|Value|Description|
+|-----|-----------|
+|`user`|CPU time spent by the process in user mode|
+|`system`|CPU time used by the system (Kernel)|
+|`guest`|CPU time used by a guest virtual machine when a hypervisor virtualizes environments|
+
+#### cpu_state
+
+The CPU states correspond the amount of time, measured in ticks, the CPU has been in specific states. These fields are measured in ticks because the underlying data from the kernel is measured in ticks (<https://docs.rs/procfs-core/latest/procfs_core/struct.CpuTime.html>) :
```
TheElectronWill, line 59: 1. The metric in Alumet is in `millisecond`: [see code](https://github.com/alumet-dev/alumet/blob/c3efb9d9f82dbe3c4c88900966e605076e442a5e/plugins/procfs/src/kernel.rs#L55-L56)
2. `cpu_state` is not an amount of time, it's an attribute that indicates the kind of cpu time that is measured

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+
+### Attributes
+
+#### Kind
+
+The kind of the memory is the allocated memory space reserved by the system or the hardware (<https://man7.org/linux/man-pages/man5/proc_pid_status.5.html>) :
+
+|Value|Description|
+|-----|-----------|
+|`resident`|Resident set size, measured in pages (same as VmRSS in `/proc/<pid>/status`)|
```
TheElectronWill, line 45: The metric in Alumet is not in pages.

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
 # Procfs plugin
 
-Collects processes- and system-related metrics by reading the `/proc` virtual filesystem.
+Collects processes and system-related metrics by reading the [proc](<https://www.kernel.org/doc/html/latest/filesystems/proc.html>) virtual filesystem on Linux based operating systems.
```
TheElectronWill, line 3:
```suggestion
Collects processes and system-related metrics by reading the [proc](https://www.kernel.org/doc/html/latest/filesystems/proc.html) virtual filesystem on Linux based operating systems.
```

`<url>` is for bare urls. When you have `[label](url)`, don't put `<>`.

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+|Name|Type|Unit|Description|Resource|ResourceConsumer|Attributes|
+|----|----|----|-----------|--------|----------------|----------|
+|`mem_total`|Gauge|bytes|Total amount of physical memory available in the system|LocalMachine|LocalMachine||
+|`mem_free`|Gauge|bytes|Refers to truly free memory, which is not used by processes, caches or buffers|LocalMachine|LocalMachine||
+|`mem_available`|Gauge|bytes|More accurate estimate of the actual available memory, taking into account free memory and also caches and buffers that can be freed|LocalMachine|LocalMachine||
+|`cached`|Gauge|bytes|Cached is the memory used for ROM cache (page cache) in memory|LocalMachine|LocalMachine||
+|`swap_cached`|Gauge|bytes|Memory that was placed in swap and brought back to RAM, but remains in swap file|LocalMachine|LocalMachine||
+|`active`|Gauge|bytes|Memory that has been used recently|LocalMachine|LocalMachine||
+|`inactive`|Gauge|bytes|Used memory that can be reclaimed, freed, or moved to swap.|LocalMachine|LocalMachine||
+|`mapped`|Gauge|bytes|Memory used for files associated with processes in memory, file and memory mapping.|LocalMachine|LocalMachine||
```
TheElectronWill, line 35: If this is about `plugins.procfs.memory.metrics`, this table lacks some values. But don't copy the whole doc, it's better to remove the table and explain that the metrics come from `/proc/meminfo`, with their names turned into snake_case (e.g. `MemTotal` becomes `mem_total`). Add a link to https://man7.org/linux/man-pages/man5/proc_meminfo.5.html.

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+|`idle`|Ticks spent in the idle state|
+|`irq`|Ticks servicing interrupts|
+|`softirq`|Ticks servicing soft interrupts|
+|`steal`|Ticks of stolen time. Stolen time is the time spent in other operating systems when running in a virtualized environment.|
+|`guest`|Ticks spent running a virtual CPU for guest operating systems under control of the linux kernel|
+|`guest_nice`|Ticks spent running a niced guest|
+
+## Configuration
+
+Here is a configuration example of the plugin. It's part of the ALUMET configuration file (eg: `alumet-config.toml`). Its configuration is composed of different sections and each of them can be enabled or disabled with the `enabled` parameter to be set to true or false.
```
TheElectronWill, line 75:
```suggestion
Here is a configuration example of the plugin. It is composed of different sections. Each section can be enabled or disabled with the `enabled` boolean parameter.
```

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+
+To active the plugin to collect metrics relative to the kernel utilization :
```
TheElectronWill, line 77: Can you create one markdown subsection for each sub-config? `### Kernel metrics` for instance

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+[plugins.procfs.processes]
+# `true` to enable the monitoring of processes.
+enabled = true
+# Watcher refresh interval.
+refresh_interval = "2s"
+# `true` to watch for new processes, `false` to only react to ALUMET events.
+strategy = "watcher"
+```
+
+Also, you can monitor groups of processes, i.e. to the subdirectories in `/proc`, each subdirectory representing a process identified by its process ID, and collect metrics about them. In the group configuration, you can specify the PID or PPID to only monitor a process that has this PID or parent PID, and a character pattern to monitor the processes whose executable path matches this regex :
```
TheElectronWill, line 123: The user does not care how `/proc` is organized exactly IMO.

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+
+```toml
+[[plugins.procfs.processes.groups]]
+# Only monitor the processes whose executable path matches this regex (No precision fixed in this case).
```
TheElectronWill, line 127: I think that `fix` is not the right verb to use here. What do you want to say?

plugins/procfs/README.md
```diff
@@ -1,3 +1,153 @@
...
+poll_interval = "2s"
+# How frequently should the processes information be flushed to the rest of the pipeline.
+flush_interval = "4s"
+```
+
+## More information
+
+### Procfs Access
+
+To grant the required access to retrieve all metrics properly by reading `/proc` filesystem, you need to configure the parameter [hidepid](<https://docs.kernel.org/filesystems/proc.html#mount-options>) by editing the configuration file `/proc/mounts`. This setting is a mount option for the `/proc` filesystem, that is used to control the visibility of processes to unprivileged users. In this way, it can define the access restriction to `/proc/<pid>/` directories, and therefore visibility of processes stats.
```
TheElectronWill, line 139: So, after investigating, is there a link to some capabilities or not?

TheElectronWill (changes requested):

PR #263, opened 2025-08-27
feat(core): implement FromStr for PrefixedUnit to support parsing any metric

### LINKED ISSUE #258: Feature: Implement FromStr for PrefixedUnit to support parsing any metric
### **Context**
This was discussed in PR #237.

Currently, `PrefixedUnit` lacks a `FromStr` implementation, which prevents parsing arbitrary metric strings (e.g., "kW", "mA") directly. This limits the ability to handle **any metric** dynamically, including those used in kwollect and other contexts. Alumet core already provides `FromStr` for `Unit` and `UnitPrefix`, so extending this to `PrefixedUnit` would be a natural and valuable addition.

### **Proposed Solution**
Implement `FromStr` for `PrefixedUnit` by combining the existing `FromStr` implementations for `Unit` and `UnitPrefix`. The implementation would:
1. Split the input string into a prefix and a unit (for example "kW" -> "k" and "W").
2. Parse the prefix and unit separately.
3. Return a `PrefixedUnit` if both parsing steps succeed.

millillitre:
# Context

This was discussed in PR #237 
This PR will resolves #258 

Currently, PrefixedUnit lacks a FromStr implementation, which prevents parsing arbitrary metric strings (e.g., "kW", "mA") directly. This limits the ability to handle any metric dynamically, including those used in kwollect and other contexts. Alumet core already provides FromStr for Unit and UnitPrefix, so extending this to PrefixedUnit would be a natural and valuable addition.

# Proposed Solution

Implement FromStr for PrefixedUnit by combining the existing FromStr implementations for Unit and UnitPrefix.

### DIFF
--- a/core/alumet/src/units.rs
+++ b/core/alumet/src/units.rs
@@ -238,6 +238,30 @@ impl Display for PrefixedUnit {
     }
 }
 
+impl FromStr for PrefixedUnit {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // List of known prefixes, sorted from longest to shortest to avoid ambiguity.
+        // Each entry is a prefix_str
+        let prefixes = [
+            "giga", "mega", "kilo", "milli", "micro", "nano", "G", "M", "k", "m", "μ", "n", "",
+        ];
+
+        // Try to find a valid prefix
+        for prefix_str in prefixes {
+            if s.starts_with(prefix_str) {
+                let unit_str = &s[prefix_str.len()..];
+                let prefix = UnitPrefix::from_str(prefix_str)?;
+                let base_unit = Unit::from_str(unit_str)?;
+                return Ok(PrefixedUnit { base_unit, prefix });
+            }
+        }
+
+        Err(anyhow!("Unknown prefix or invalid unit in '{}'", s))
+    }
+}
+
 impl UnitPrefix {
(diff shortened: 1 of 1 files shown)


core/alumet/src/units.rs
```diff
@@ -335,4 +359,22 @@ mod tests {
         assert_eq!(parse_self(UnitPrefix::Mega), UnitPrefix::Mega);
         assert_eq!(parse_self(UnitPrefix::Giga), UnitPrefix::Giga);
     }
+
+    #[test]
+    fn prefixed_unit_serde() {
```
TheElectronWill, line 364: Can you add some tests with bad inputs, to assert that an error is properly returned?

core/alumet/src/units.rs
```diff
@@ -238,6 +238,30 @@ impl Display for PrefixedUnit {
     }
 }
 
+impl FromStr for PrefixedUnit {
+    type Err = anyhow::Error;
```
TheElectronWill, line 242: Let's create a dedicated error type. You can use [`thiserror`](https://crates.io/crates/thiserror) to make it easy.

TheElectronWill (changes requested):
Thanks for the PR.

core/alumet/src/units.rs
millillitre: I used the `assert!(result.is_err())` to check that parsing fails for some examples of invalid inputs, is it what you were asking for? Or do you want it to verify the specific error message?

TheElectronWill (approved):

PR #268, opened 2025-08-29
feat(slurm): Is it a job or a step

### LINKED ISSUE #266: Differenciate the job and its steps in the Slurm plugin
I would like to know if the measure returned is the total job cpu usage of just a step inside. It's currently possible to do it after using the `consumer_id` 

I would like a attribute which indicate precisely if it's the main job or a substep

AngeCyp:
add the specification if it's a job or a step in slurm plugin.
It returns in attribute: 

- `job` if it's the main job
- `step` if it's a step of the job

Fix #266

### DIFF
--- a/plugins/cgroups/slurm/src/source.rs
+++ b/plugins/cgroups/slurm/src/source.rs
@@ -1,7 +1,7 @@
-use alumet::{measurement::AttributeValue, pipeline::elements::source::trigger::TriggerSpec};
+use alumet::pipeline::elements::source::trigger::TriggerSpec;
 use util_cgroups::Cgroup;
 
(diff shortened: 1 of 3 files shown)


plugins/cgroups/slurm/src/source.rs
```diff
@@ -49,6 +51,16 @@ impl CgroupSetupCallback for JobSourceSetup {
             let job_id = find_jobid_in_attrs(&attrs).expect("job_id should be set");
             // give a nice name
             name = format!("slurm-job-{}", job_id);
+
+            // Check if it's the job or a step
+            let end_pattern = format!("job_{}", job_id);
+            if cgroup.canonical_path().ends_with(&end_pattern) {
+                // It's the job
+                attrs.push(("represent".to_string(), AttributeValue::Str("job")));
```
TheElectronWill, line 59: What do you think about `job_part` instead of "represent"?

PR #273, opened 2025-09-03
test(plugin-csv): add unit and integration tests for plugin csv

Aymeric-Brochier:


plugins/csv/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 mod csv;
-mod output;
+pub mod output;
```
TheElectronWill, line 2: This should not be made public

plugins/csv/src/lib.rs
```diff
@@ -79,3 +79,50 @@ impl Default for Config {
...
+        let agent = agent::Builder::from_pipeline(plugins, pipeline)
+            .build_and_start()
+            .expect("startup failure");
+
+        let handle = agent.pipeline.control_handle();
+        // Force shutdown
+        handle.shutdown();
+        agent
+            .wait_for_shutdown(Duration::new(5, 0))
+            .expect("error while running");
```
TheElectronWill, line 126: Calling shutdown just after `build_and_start` and just before `wait_for_shutdown` is usually not a good idea: the pipeline will not run.

plugins/csv/src/lib.rs
```diff
@@ -79,3 +79,50 @@ impl Default for Config {
...
+    };
+    use std::time::Duration;
+
+    fn config_to_toml_table(config: &Config) -> toml::Table {
+        toml::Value::try_from(config).unwrap().as_table().unwrap().clone()
+    }
+
+    #[test]
+    fn start_stop() {
+        let default_config = Config { ..Default::default() };
```
TheElectronWill, line 103: What is this test testing? The integration test is good enough: it already does start and stop.

plugins/csv/src/output.rs
```diff
@@ -188,3 +188,57 @@ impl alumet::pipeline::Output for CsvOutput {
...
+mod tests {
+
+    use alumet::{
+        measurement::{MeasurementBuffer, MeasurementPoint, Timestamp, WrappedMeasurementValue},
+        metrics::RawMetricId,
+        resources::{Resource, ResourceConsumer},
+    };
+    use std::{collections::HashSet, time::UNIX_EPOCH};
+
+    use crate::output::{collect_attribute_keys, escape_late_attribute};
```
TheElectronWill, line 202: These functions are in the parent module. Use `super::` to access them.

plugins/csv/tests/output.rs
```diff
@@ -0,0 +1,276 @@
...
+                        expected_string_no_late_attributes
+                    );
+                }
+            },
+        )
+        // Adding another point to have late attributes
+        .test_output(
+            elastic_output.clone(),
+            {
+                move |ctx: &mut OutputCheckInputContext| -> MeasurementBuffer {
```
TheElectronWill, line 131: Can this be written
```suggestion
                move |ctx| {
```
?

plugins/csv/tests/output.rs
```diff
@@ -0,0 +1,276 @@
...
+fn csv_output() {
+    let _ = env_logger::Builder::from_default_env().try_init();
+
+    // Prepare the plugin
+
+    let tmp = tempfile::tempdir().unwrap();
+    let file_path = tmp.path().join("alumet-output.csv");
+
+    let result_file: String = file_path.clone().into_os_string().into_string().unwrap();
+    let result_file_2 = result_file.clone();
```
TheElectronWill, line 62: You don't need a string to use `std::fs::read_to_string`. You can keep a `Path` or `PathBuf`.

plugins/csv/tests/output.rs
```diff
@@ -0,0 +1,276 @@
...
+                    // Give input to the pipeline
+                    MeasurementBuffer::from(vec![point1])
+                }
+            },
+            {
+                move || {
+                    assert_eq!(
+                        fs::read_to_string(result_file_2.as_str()).unwrap(),
+                        expected_string_with_late_attributes
+                    );
```
TheElectronWill, line 146: good! Here we clearly see the input and the output, side by side :)

PR #278, opened 2025-09-09
test(plugin-elasticsearch): add tests for plugin elasticsearch

Aymeric-Brochier:


plugins/elasticsearch/src/plugin.rs
```diff
@@ -170,7 +170,68 @@ mod config {
...
+    use crate::{ElasticSearchPlugin, api, plugin::config::AuthConfig};
+
+    use super::config::Config;
+
+    fn config_to_toml_table(config: &Config) -> toml::Table {
+        toml::Value::try_from(config).unwrap().as_table().unwrap().clone()
+    }
+
+    #[test]
+    fn start_stop() {
```
TheElectronWill, line 195: not needed

plugins/elasticsearch/src/plugin.rs
```diff
@@ -211,4 +272,48 @@ mod tests {
...
         assert_eq!(parsed.metric_unit_as_index_suffix, true);
     }
+
+    #[test]
+    fn try_from_basic_file_api_key_bearer() {
+        // Basic File
+
+        let tmp = tempfile::tempdir().unwrap();
+        let file_path = tmp.path().join("basicfile.csv");
+        fs::write(file_path.clone(), "bob:very_secure").unwrap();
```
TheElectronWill, line 282: Instead of cloning, use `&file_path` when you can

plugins/elasticsearch/src/plugin.rs
```diff
@@ -170,7 +170,68 @@ mod config {
...
+        // Build and start the agent
+        let agent = agent::Builder::from_pipeline(plugins, pipeline)
+            .build_and_start()
+            .expect("startup failure");
+
+        let handle = agent.pipeline.control_handle();
+        // Force shutdown
+        handle.shutdown();
+        agent
+            .wait_for_shutdown(Duration::new(5, 0))
```
TheElectronWill, line 231: use `Duration::from_secs` so the unit is explicit

TheElectronWill (changes requested):

TheElectronWill (approved):

PR #314, opened 2025-11-20
feat(amd-gpu): AMD GPU plugin integration + tests

victoryeagle77:


plugins/amd-gpu/src/interface.rs
```diff
@@ -0,0 +1,470 @@
...
+        amdsmi_socket_handle, amdsmi_status_t, amdsmi_temperature_metric_t, amdsmi_temperature_type_t,
+        amdsmi_voltage_metric_t, amdsmi_voltage_type_t, libamd_smi,
+    },
+};
+
+/// Error treatment concerning AMD SMI library.
+///
+/// # Arguments
+///
+/// Take a status of [`amdsmi_status_t`] provided by AMD SMI library to catch dynamically the occurred error.
```
TheElectronWill: It's good to document functions and types thoroughly, but 1. the `Arguments` section is usually not needed (it's standard not to use it in Rust) and 2. Here, it doesn't bring additional value: we can see that the field is of type `amdsmi_status_t` and that it's an error given by amdsmi ^^

plugins/amd-gpu/src/interface.rs
```diff
@@ -0,0 +1,470 @@
...
+        } else {
+            Err(AmdError(result))
+        }
+    }
+
+    /// Call the unsafe C binding function [`amdsmi_get_gpu_activity`] to retrieve gpu activity values.
+    ///
+    /// # Arguments
+    ///
+    /// Address pointer on a AMD GPU device coming from [`ProcessorHandle`].
```
TheElectronWill: This is the doc of the amdsmi_get_gpu_activity, right? Don't put it here, here it's the documentation of the Rust function `get_device_activity`

TheElectronWill (changes requested):
A quick pass between two chapters, I've not reviewed everything. I think that the `#[test]` could be organized a bit better, grouped by what they test.

plugins/amd-gpu/src/amd/device.rs
```diff
@@ -0,0 +1,371 @@
...
+    /// Detected working features on the AMD GPU device.
+    pub working_devices: usize,
+}
+
+impl AmdGpuDevices {
+    /// Detects all AMD GPUs and returns an AmdGpuDevices object.
+    pub fn detect(amdsmi: &dyn AmdSmiLibProvider, skip_failed_devices: bool) -> anyhow::Result<Self> {
+        // Get our global AMD SMI instance
+        let mut devices = HashMap::new();
+        let mut failure_count = 0;
```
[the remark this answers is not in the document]
victoryeagle77: Ah yes, it was for old lines that no longer exist.

plugins/amd-gpu/src/interface.rs
victoryeagle77: No it's not the doc for `amdsmi_get_gpu_activity`, it's for `get_device_activity` like any other functions in the interface

TheElectronWill: Okay, so you need to update it to document the Rust function with Rust terms. For instance, the function does not take an "address pointer" and does not return "gfx".

plugins/amd-gpu/amd-smi-lib/include/amdsmi.h
TheElectronWill: this should be in a separate crate in a separate repo and PR

plugins/amd-gpu/amd-smi-lib/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+pub struct AmdEnergyConsumptionInfo {
+    /// The energy consumption value of an AMD GPU device since the last boot in micro Joules.
+    pub energy: u64,
+    /// Precision factor of the energy counter in micro Joules.
+    pub resolution: f32,
+    /// The time during which the energy value is recovered in ns.
+    pub timestamp: u64,
+}
+
+pub type MockableAmdSmi = Arc<dyn AmdSmiTrait>;
```
TheElectronWill: traits + dyn could work, but let's avoid unnecessary overhead here: we know, at compile time, if we'll use the mock or not. You should use structs and cfg annotations.

plugins/amd-gpu/amd-smi-lib/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
TheElectronWill: how do you plan to differentiate between mock/non-mock in the new crate? (see schema for the relevant thing to do with `cfg` ^^)

plugins/amd-gpu/amd-smi-lib/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// Quit amd-smi library and clean properly its resources.
+    fn stop(&self) -> Result<(), AmdError> {
+        let result = unsafe { self.amdsmi.amdsmi_shut_down() };
+        get_value(result)
+    }
+
+    /// Retrieves a set of [`SocketHandle`] structure containing socket handles associated to a GPU device.
+    fn get_socket_handles(&self) -> Result<Vec<MockableAmdSocketHandle>, AmdError> {
+        let mut socket_count = 0;
+        let result = unsafe { self.amdsmi.amdsmi_get_socket_handles(&mut socket_count, null_mut()) };
```
TheElectronWill: SAFETY comment?

TheElectronWill (changes requested):

plugins/amd-gpu/amd-smi-lib/src/lib.rs
victoryeagle77: ?

plugins/amd-gpu/amd-smi-lib/Cargo.toml
victoryeagle77: With my design, we don't need to choose what should be different because all the code is usable; it's only at the plugin level that we choose, depending on whether we use it in test or production code.

TheElectronWill: Sure, but I don't want to compile the mocks when I'm not using them.

plugins/amd-gpu/src/amd/features.rs
```diff
@@ -0,0 +1,322 @@
...
+/// Checks if a feature is supported by the available GPU by inspecting the return type of an AMDSMI function.
+pub fn is_supported<T>(res: Result<T, amdsmi_status_t>) -> Result<bool, amdsmi_status_t> {
+    match res {
+        Ok(_) => Ok(true),
+        Err(NO_PERM) => Ok(false),
+        Err(NOT_SUPPORTED) => Ok(false),
+        Err(NOT_YET_IMPLEMENTED) => Ok(false),
+        Err(UNEXPECTED_DATA) => Ok(false),
+        Err(e) => Err(e),
+    }
```
TheElectronWill:
```suggestion
    match res {
        Ok(_) => Ok(true),
        Err(NO_PERM | NOT_SUPPORTED | NOT_YET_IMPLEMENTED | UNEXPECTED_DATA) => Ok(false),
        Err(e) => Err(e),
    }
```

plugins/amd-gpu/README.md
```diff
@@ -0,0 +1,141 @@
...
+## Requirements
+
+- Linux operating system.
+- AMD GPU(s).
+- Installation of libdrm.
+- Installation of clang and libclang.
+- Installation of `amdsmi` software : https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/quick-start.html and if it doesn't work properly on your linux distribution:
+  - Installation of a recent [cmake](https://cmake.org/) version.
+  - Build `amdsmi` [source](https://github.com/ROCm/amdsmi.git) code manually with recent cmake version (otherwise the builder will fail to compile and show the version you need to have).
+- Set and configure some permissions on system to run properly all metrics: https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/prerequisites.html#configuring-permissions-for-gpu-access.
```
TheElectronWill: This needs to be updated, we don't need to build amdsmi from source anymore.

Issue #306, opened 2025-11-05
On OAR, GPU measurements should be annotated with the job id


TheElectronWill:
This is implemented for Slurm but not for OAR yet.

Symptom: energy-attribution does not work with the GPU data and OAR jobs, as reported by @ezychlagre.
It will work when the annotation-transform is implemented for OAR.

PR #315, opened 2025-12-02
Feat/add job annotation

AngeCyp:
Generalization of the job annotation transform for several cgroups plugins:

- OAR
- Slurm
- K8S

It fixes: #306 

Also add some unit tests.


plugins/cgroups/util-cgroups-plugins/tests/tests.rs
```diff
@@ -0,0 +1,480 @@
...
+    job_annotation_transform::{
+        CachedCgroupHierarchy, JobAnnotationTransform, JobTagger, OptionalSharedHierarchy, SharedCgroupHierarchy,
+    },
+};
+
+use lazy_static::lazy_static;
+
+const SYSFS_CGROUP: &str = "/sys/fs/cgroup";
+
+lazy_static! {
```
TheElectronWill: these are used in only one function, you can declare them locally (in the function) and remove the dependency

plugins/cgroups/util-cgroups-plugins/Cargo.toml
```diff
@@ -9,6 +9,7 @@ description = "Common code to create cgroup-based Alumet plugins."
 alumet.workspace = true
 anyhow.workspace = true
 humantime-serde.workspace = true
+lazy_static = "1.5.0"
```
TheElectronWill: should not be a non-dev dependency

TheElectronWill (approved):

PR #322, opened 2025-12-18
Plugin NVML: Allow to choose between a full mode and minimal mode

TheElectronWill:
Validated on G5K kinovis, NVIDIA L40S.


titouanj (approved):
need rebase

PR #335, opened 2026-01-28
Fix plugin ordering

TheElectronWill:
The order in which the plugins are initialised now correspond to the configuration file. This should have done sooner and we almost thought that it was the case… Now, it's _actually_ implemented for real!

This helps running the tdp + the attribution together.


guigomcha (approved):
LGTM

PR #340, opened 2026-02-03
fix(prometheus-exporter): sanitize correctly units and metric's names

### LINKED ISSUE #316: Prometheus fails to read nvml-exporter metrics because of invalid characters
**Describe the bug**
The nvml-exporter exports metrics like "nvml_sm_utilization_alumet_%" "nvml_encoder_sampling_period_alumet_μs" which contains characters that are invalid for prometheus. Have a look at the documentation here: https://prometheus.io/docs/concepts/data_model/. It says there:

> Metric names SHOULD match the regex [a-zA-Z_:][a-zA-Z0-9_:]* for the best experience and compatibility (see the warning below). Metric names outside of that set will require quoting e.g. when used in PromQL (see the [UTF-8 guide](https://prometheus.io/docs/guides/utf8/#querying)).

Prometheus returns me an error "Error scraping target: expected text in HELP"

Perhaps it would be better to write out all of the non-compatible characters explicitly, e.g. "%" -> "percentage", etc...

**Environment information**
OS: ubuntu 24.04
Alumet version²: 0.9.1
Enabled plugins: nvml, prometheus-exporter

²: please include the commit hash if it's not a release

**To Reproduce**
1. Run alumet agent with both prometheus-exporter and nvml plugins enabled
2. try to read the endpoint "/metrics" of the alumet-agent with prometheus
3. See error Error scraping target: expected text in HELP

**Expected behavior**
Prometheus reads metrics endpoint with nvml metrics just fine.

**Logs / Output**

```
# HELP nvml_sm_utilization_alumet_% Utilization of the GPU streaming multiprocessors by the process.
# TYPE nvml_sm_utilization_alumet_% gauge
# UNIT nvml_sm_utilization_alumet_% %
nvml_sm_utilization_alumet_%{resource_consumer_id="2908708",resource_consumer_kind="process",resource_id="00000000:81:00.0",resource_kind="gpu"} 9.0
nvml_sm_utilization_alumet_%{resource_consumer_id="15465",resource_consumer_kind="process",resource_id="00000000:81:00.0",resource_kind="gpu"} 1.0
```

titouanj:
Fixes #316

### DIFF
--- a/plugins/prometheus-exporter/src/output.rs
+++ b/plugins/prometheus-exporter/src/output.rs
@@ -24,7 +24,6 @@ pub struct MetricState {
 #[derive(Clone)]
 pub struct PrometheusOutput {
     pub state: MetricState,
-    use_unit_display_name: bool,
     add_attributes_to_labels: bool,
     prefix: String,
     suffix: String,
@@ -33,7 +32,6 @@ pub struct PrometheusOutput {
 
 impl PrometheusOutput {
     pub fn new(
-        use_unit_display_name: bool,
         add_attributes_to_labels: bool,
         port: u16,
         host: String,
@@ -52,7 +50,6 @@ impl PrometheusOutput {
 
         Ok(Self {
             state,
-            use_unit_display_name,
             add_attributes_to_labels,
             prefix,
             suffix,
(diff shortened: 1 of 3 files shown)


plugins/prometheus-exporter/src/output.rs
```diff
@@ -137,17 +134,56 @@ impl alumet::pipeline::Output for PrometheusOutput {
...
 }
 
-fn get_unit_string(full_metric: &Metric, use_unit_display_name: bool) -> String {
-    if use_unit_display_name {
-        full_metric.unit.display_name()
-    } else {
-        full_metric.unit.unique_name()
+/// Helper function that returns the metric's unit according to the Prometheus
+/// [base units](https://prometheus.io/docs/practices/naming/#base-units) documentation.
+fn get_unit_string(full_metric: &Metric) -> String {
```
TheElectronWill: Could you add unit tests for this function, e.g. with `milli(Joule)` and `Percent` ?

plugins/prometheus-exporter/src/output.rs
```diff
@@ -128,7 +125,7 @@ impl alumet::pipeline::Output for PrometheusOutput {
             // Update metric value
             let gauge = family.get_or_create(&labels);
             match m.value {
```
TheElectronWill: `m.value.as_f64()` if it works

guigomcha (approved):
LGTM

PR #342, opened 2026-02-06
[plugin procfs] More accurate memory stats (and a bugfix)

### LINKED ISSUE #327: [plugin procfs] Better accuracy for memory measurement
Firstly thank you for developing Alumet!

One of my colleagues recently tried to use the `procfs` plugin to perform a black-box memory profiling of a research code. 

We found the values to be surprisingly quite off from expected.

On going through the [source](https://github.com/alumet-dev/alumet/blob/58901d6024c9f3a1f236e3a86763991481bbbfff/plugins/procfs/src/process.rs#L69C1-L69C75) we found that the reported values for the consumer `process` are taken from `/proc/pid/statm`. 

I am no Linux expert, but according to [proc_pid_statm](https://man7.org/linux/man-pages/man5/proc_pid_statm.5.html), these measured memory values are inaccurate. For accurate measurements we should be using `Pss` value from [proc_pid_smaps](https://man7.org/linux/man-pages/man5/proc_pid_smaps.5.html), where it states
>  ... the amount of the
              mapping that is currently resident in RAM ("Rss"), the
              process's proportional share of this mapping ("Pss"), the
              number of clean and dirty shared pages in the mapping, and
              the number of clean and dirty private pages in the mapping....

Hope this can be implemented soon :)

TheElectronWill:
Resolves #327

### DIFF
--- a/plugins/procfs/src/process.rs
+++ b/plugins/procfs/src/process.rs
@@ -25,6 +25,14 @@ use alumet::{
 use anyhow::Context;
 use procfs::{self, FromRead, ProcError, process::Process};
 use regex::Regex;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MemoryStatsMode {
+    Quick,
+    SlowButAccurate,
+}
 
 /// Reads process stats from `/proc/<pid>` for some `<pid>`.
 struct ProcessStatsProbe {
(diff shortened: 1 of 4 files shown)


plugins/procfs/src/process.rs
```diff
@@ -98,15 +116,57 @@ fn stop_if_io_not_found(err: io::Error) -> PollError {
...
+                    procfs::process::SmapsRollup::from_read(&mut self.reader_mem).map_err(stop_if_proc_not_found)?;
+                let smaps_rollup = &smaps_rollup
+                    .memory_map_rollup
+                    .0
+                    .first()
+                    .context("smaps_rollup returned no data")?
+                    .extension
+                    .map;
+                let rss = smaps_rollup
+                    .get("Rss")
```
thealanjason: Probably a naive question, but shouldn't we be using "Pss" value instead, since we're interested in the process specific memory usage? I understand that the value in `Rss` may be shared across multiple processes. Quoting the [proc_pid_smaps](https://man7.org/linux/man-pages/man5/proc_pid_smaps.5.html) again:

```
              lines show the size of the mapping, the amount of the
              mapping that is currently resident in RAM ("Rss"), the
              process's proportional share of this mapping ("Pss"),
```

PR #343, opened 2026-02-09
test(plugins/rapl): Add tests

victoryeagle77:


TheElectronWill (changes requested):
Hi!
1. The pr name and commit message are inconsistent. Is it a feature, a test or a fix? What did you fix? I think it should be `tests(plugins/rapl): <description>`
2. All the changes are combined in a single commit, but that's a mixup of moves, refactoring, new tests, etc. To make it easier to review, you should separate these in at least two commits.
3. We should discuss about the organisation of tests. Usually, unit tests go in the file that they test (I know, there was a `test_mock` aside here).

PR #377, opened 2026-03-31
test(plugins/rapl): Updates to add 80% test code coverage for RAPL plugin

victoryeagle77:
You can read this PR commit by commit for a more easier review


plugins/rapl/src/tests/tests.rs
```diff
@@ -0,0 +1,308 @@
...
+    units::Unit,
+};
+use std::thread;
+use std::{
+    path::{Path, PathBuf},
+    time::Duration,
+};
+use tempfile::tempdir;
+
+#[cfg(test)]
```
TheElectronWill: the whole module is already in cfg(test)

plugins/rapl/src/tests/tests.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    thread::sleep(Duration::from_millis(200)); // don't shutdown right away, else we get in trouble
+    agent.pipeline.control_handle().shutdown();
+    agent.wait_for_shutdown(Duration::from_secs(10)).unwrap();
+
+    Ok(())
+}
+
+#[test]
+fn test_runtime_with_powercap() -> anyhow::Result<()> {
```
TheElectronWill, line 30: You can combine the startup and runtime tests in one agent run.
`.with_expectations` can be called multiple times

plugins/rapl/src/tests/tests.rs
```diff
@@ -0,0 +1,308 @@
...
+        .expect_metric::<f64>("rapl_consumed_energy", Unit::Joule)
+        .expect_source("rapl", "in");
+
+    let agent = Builder::new(plugins)
+        .with_expectations(startup_expectations)
+        .build_and_start()
+        .unwrap();
+
+    agent.pipeline.control_handle().shutdown();
+    agent.wait_for_shutdown(Duration::from_secs(10)).unwrap();
```
TheElectronWill: 10s is too long, it should shutdown in 1s or so

plugins/rapl/src/tests/tests.rs
```diff
@@ -0,0 +1,308 @@
...
+fn test_with_powercap_only() -> anyhow::Result<()> {
+    let mut plugins = PluginSet::new();
+
+    let tmp = create_valid_powercap_mock()?;
+    let base_path = tmp.path().to_owned();
+
+    let source_config = Config {
+        poll_interval: Duration::from_secs(1),
+        flush_interval: Duration::from_secs(1),
+        no_perf_events: false, // Enable perf_events at starting to switch in powercap
```
TheElectronWill: the comment doesn't make much sense here, are you trying to say "disable perf_events to force powercap` ?

plugins/rapl/src/tests/tests.rs
```diff
@@ -0,0 +1,308 @@
...
+                assert_eq!(measurement.value, WrappedMeasurementValue::F64(30000.0));
+            },
+        );
+
+    let agent = agent::Builder::new(plugins)
+        .with_expectations(runtime_expectations)
+        .build_and_start()
+        .unwrap();
+
+    agent.wait_for_shutdown(Duration::from_secs(10)).unwrap();
```
TheElectronWill: put the timeout in a global var

TheElectronWill (changes requested):

plugins/rapl/src/tests/tests.rs
victoryeagle77: Yes you're right 1s is better

plugins/rapl/src/tests/tests.rs
victoryeagle77: Yes I forget to remove it.

plugins/rapl/src/tests/tests.rs
victoryeagle77: Yes right it's more revealing

plugins/rapl/src/tests/tests.rs
victoryeagle77: It was not initially my tests but it's right we can combine them.

plugins/rapl/src/perf_event.rs
```diff
@@ -543,4 +565,136 @@ mod tests {
...
+
+        let result = expected_power_events.perf_event_open(32, 0);
+        match result {
+            // OS permissions are enough to pen perf_event
+            Ok(fd) => {
+                assert!(fd >= 0);
+            }
+            // Permission denied
+            Err(e) => {
+                println!("{e}");
```
TheElectronWill: What are you testing here exactly?

plugins/rapl/src/perf_event.rs
```diff
@@ -543,4 +565,136 @@ mod tests {
...
+            domain: RaplDomainType::Package,
+            code: 1,
+            unit: "Joules".into(),
+            scale: SCALE,
+        };
+
+        let (fd, _write_fd) = fake_fd();
+        let actual_power_events = move |_e: &PowerEvent, _pmu, _cpu| Ok(fd);
+
+        let result = expected_power_events.open_power_event(1, 0, 0, actual_power_events);
```
TheElectronWill: Naming is a bit weird, usually you do `assert_eq(actual, expected)`, here the `move |…| Ok(fd)` is a close that simulates perf_event_open.

plugins/rapl/src/perf_event.rs
```diff
@@ -543,4 +565,136 @@ mod tests {
...
+        assert!(result.is_ok());
+
+        let opened_power_event = result.unwrap();
+        assert_eq!(opened_power_event.scale, SCALE.into());
+        assert_eq!(opened_power_event.domain, expected_power_events.domain);
+        assert_eq!(opened_power_event.resource, expected_power_events.domain.to_resource(0));
```
TheElectronWill:
```suggestion
        let event = PowerEvent {
            name: "pkg".into(),
            domain: RaplDomainType::Package,
            code: 1,
            unit: "Joules".into(),
            scale: SCALE,
        };

        let (fd, _write_fd) = fake_fd();
        let fake_open = move |_e: &PowerEvent, _pmu, _cpu| Ok(fd);

        let result = event.open_power_event(1, 0, 0, fake_open);
        let opened_power_event = result.expect("should not fail");
        assert_eq!(opened_power_event.scale, SCALE.into());
        assert_eq!(opened_power_event.domain, event.domain);
        assert_eq!(opened_power_event.resource, event.domain.to_resource(0));
```

TheElectronWill: The closure is not the "expected" event, therefore I've renamed the variables a little bit.

plugins/rapl/src/perf_event.rs
```diff
@@ -543,4 +565,136 @@ mod tests {
...
+
+        // Initialize counters from first read
+        write(&fd_1, &42u64.to_ne_bytes())?;
+        write(&fd_2, &43u64.to_ne_bytes())?;
+
+        evt_1.read_counter_diff()?;
+        evt_2.read_counter_diff()?;
+
+        let mut probe = PerfEventProbe {
+            metric: unsafe { zeroed() },
```
TheElectronWill: Never do this for a metric id: there is no guarantee (from Alumet) that 0 is a valid representation of an id.
To use metrics, use the Alumet testing API.

plugins/rapl/src/powercap.rs
```diff
@@ -813,4 +829,118 @@ mod tests {
...
+                entry_type: File("10"),
+            },
+        ];
+
+        create_mock_layout(base_path, &entries)?;
+
+        let result = all_power_zones_from_path(base_path);
+
+        assert!(result.is_err());
+        assert!(result.unwrap_err().to_string().contains("Unknown RAPL powercap zone"));
```
TheElectronWill: You don't need to assert `is_err()` when you `unwrap_err()` just after.
The unwrap is enough.

TheElectronWill (changes requested):

plugins/rapl/src/powercap.rs
victoryeagle77: Yes it's true, I forgot to remove it because I decided after to make with the unwrap

plugins/rapl/src/perf_event.rs
victoryeagle77: Yes it's true it was not very well named

plugins/rapl/src/perf_event.rs
victoryeagle77: Yes it's true it was not very well named

plugins/rapl/src/perf_event.rs
victoryeagle77: Yes the test testing `perf_event_open` but the matches branches are useless, it's already cover when we test here : expected_power_events.perf_event_open(32, 0)
But it's just a syscall smoke test that don't check the function logic, to avoid to extract again in mockable logic, but I can do that If you prefer

plugins/rapl/src/powercap.rs
```diff
@@ -813,4 +829,116 @@ mod tests {
...
+            Entry {
+                path: "intel-rapl:0/energy_uj",
+                entry_type: File("10"),
+            },
+        ];
+        create_mock_layout(base_path, &entries)?;
+
+        let zones = all_power_zones_from_path(base_path)?.flat;
+
+        let metric = unsafe { zeroed() };
```
TheElectronWill: You have another `unsafe { zeroed() };` here. You should find another way to write this test.

victoryeagle77: Yes I don't see that, moreover this test was not extremly usefull in functionnal and coverage words so we can remove it.

plugins/rapl/src/perf_event.rs
victoryeagle77: Yes I don't see that, moreover this test was not extremly usefull in functionnal and coverage words so we can remove it.

TheElectronWill (approved):

PR #382, opened 2026-04-02
feat: Add install.sh to the repo

Fyspeed:
This script automatically installs the version of alumet-agent you want for your distribution and architecture

- closes https://github.com/alumet-dev/alumet/issues/368

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,381 @@
+#!/bin/sh
+set -e
+
+usage() {
+  this=$1
+  cat <<EOF
+$this: download alumet-agent from ${OWNER}/${REPO}
+
+Usage: $this [-d] [<tag>]
+  -d turns on debug logging
+   <tag> is a tag from
+   https://github.com/${OWNER}/${REPO}/releases
+   If tag is missing, then the latest will be used.
+
+EOF
+  exit 2
+}
+parse_args() {
+  while getopts "b:dh?x" arg; do
+    case "$arg" in
+      d) log_set_priority 10 ;;
+      h | \?) usage "$0" ;;
+      x) set -x ;;
+    esac
+  done
+  shift $((OPTIND - 1))
+  TAG=$1
+}
+execute() {
+  tmpdir=$(mktemp -d)
+  http_download "${tmpdir}/${PACKAGE_ID}" "${PACKAGE_URL}"
+  verify_hash "${tmpdir}/${PACKAGE_ID}"
+  case $DISTRIB in
+    ubuntu*|debian*) sudo apt-get install -yq --allow-downgrades "${tmpdir}/${PACKAGE_ID}" > "/dev/null";;
+    *)  yum install -yq "${tmpdir}/${PACKAGE_ID}" > "/dev/null";;
+  esac
+  log_info "Installed Alumet successfully"
+  rm -rf "${tmpdir}"
+}
+tag_to_version() {
+  if [ -z "${TAG}" ]; then
+    log_info "Checking GitHub for latest tag"
+  else
+    log_info "Checking GitHub for tag '${TAG}'"
+  fi
+
+  REALTAG=$(github_release "$OWNER/$REPO" "${TAG}") && true
+  if test -z "$REALTAG"; then
+    log_crit "Unable to find '${TAG}' - use 'latest' or see https://github.com/${OWNER}/${REPO}/releases for details"
+    exit 1
+  fi
+  # if version starts with 'v', remove it
+  TAG="$REALTAG"
+  VERSION=${TAG#v}
+}
+verify_hash() {
+  package_hashed=$(hash_sha256 "$1")
+  log_debug "Comparing digests expected ${CHECKSUM} vs obtained ${package_hashed}"
+  if [ "$package_hashed" != "$CHECKSUM" ]; then
+    log_err "hash_sha256_verify checksum for '$1' did not verify ${package_hashed} vs $CHECKSUM"
+    return 1
+  fi
+}
+log_prefix() {
+	echo "$PREFIX"
+}
+
+cat /dev/null <<EOF
+------------------------------------------------------------------------
+https://github.com/client9/shlib - portable posix shell functions
+Public domain - http://unlicense.org
+https://github.com/client9/shlib/blob/HEAD/LICENSE.md
+but credit (and pull requests) appreciated.
+------------------------------------------------------------------------
+EOF
+is_command() {
+  command -v "$1" >/dev/null
+}
+echoerr() {
+  echo "$@" 1>&2
+}
+_logp=6
+log_set_priority() {
+  _logp="$1"
+}
+log_priority() {
+  if test -z "$1"; then
+    echo "$_logp"
+    return
+  fi
+  [ "$1" -le "$_logp" ]
+}
+log_tag() {
+  case $1 in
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -0,0 +1,381 @@
...
+set -e
+
+usage() {
+  this=$1
+  cat <<EOF
+$this: download alumet-agent from ${OWNER}/${REPO}
+
+Usage: $this [-d] [<tag>]
+  -d turns on debug logging
+   <tag> is a tag from
```
ahuret: I would use a flag as tag is optional (`-t` ?)

install.sh
```diff
@@ -0,0 +1,381 @@
...
+  log_debug "Checksum found: $CHECKSUM"
+  log_debug "File found: $PACKAGE_URL"
+}
+
+OWNER="alumet-dev"
+REPO="alumet"
+PREFIX="$OWNER/$REPO"
+
+AGENT="alumet-agent"
+OS=$(uname_os)
```
ahuret: looks like `OS` is unused, is it to prepare the future ?

ahuret (changes requested):

install.sh
Fyspeed: I removed the variable, and renamed uname_os -> check_os
In the future if other OS are supported we can change it back

install.sh
titouanj: Don't agree, for me it's like when doing:
podman pull hello-world => use latest 
podman pull hello-world:1.2.3

TheElectronWill: For explicitness, and coherence if we add a ` --local` flag later, I propose to use `--version` or `--release` to choose the version to install. Btw, using "tag" is a bit weird, because only releases have a pre-built package.

Fyspeed: I've added a -l flag to install to ~/.local/bin, and added a -r flag to chose a specific release to install
Also if you install locally the binary will be named alumet-agent-local instead of alumet-agent (so that if you have both installed, you can still differentiate them)

install.sh
```diff
@@ -0,0 +1,443 @@
+#!/bin/sh
+set -e
+
+usage() {
+  this=$1
+  cat <<EOF
+$this: download alumet-agent from ${OWNER}/${REPO}
+
+Usage: $this [-d] [-l] [-r <release>] [-v <distrib version>]
```
titouanj: v for verbose, and now d is available for distrib version
```suggestion
Usage: $this [-v] [-l] [-r <release>] [-d <distrib version>]
```

install.sh
```diff
@@ -0,0 +1,443 @@
...
+  PACKAGE_URL=$(echo "$res" | cut -d ' ' -f 1) 
+  CHECKSUM=$(echo "$res" | cut -d ' ' -f 2  | awk '{gsub(/^.{7}/,"");}1' )
+
+  log_debug "Checksum found: $CHECKSUM"
+  log_debug "File found: $PACKAGE_URL"
+
+  if [ -z "$PACKAGE_URL" ]; then
+    log_err "No package found matching your distribution version (version ${VERSION} of distrib ${DISTRIB}).
+      Go to https://github.com/${OWNER}/${REPO}/releases to see if a matching package exists.
+      You may also use the -v flag to download a package available on a previous version of your distribution."
```
titouanj:
```suggestion
      You may also use the -d flag to download a package available on a previous version of your distribution."
```

install.sh
```diff
@@ -0,0 +1,443 @@
...
+  log_debug "File found: $PACKAGE_URL"
+
+  if [ -z "$PACKAGE_URL" ]; then
+    log_err "No package found matching your distribution version (version ${VERSION} of distrib ${DISTRIB}).
+      Go to https://github.com/${OWNER}/${REPO}/releases to see if a matching package exists.
+      You may also use the -d flag to download a package available on a previous version of your distribution."
+    return 1
+  fi
+}
+
```
titouanj: These should be factorized. The only difference should be wget vs curl, but same awk, logs and other stuff.
Currently, L372-373 are different than L408-410

install.sh
```diff
@@ -0,0 +1,443 @@
...
+    /"digest"/ { digest = $4 }
+    
+    # When all three fields are fullfilled
+    (asset_name && url && digest) {
+        # Check for pattern
+        if (asset_name ~ pattern_dot || asset_name ~ pattern_underscore) {
+          print url, digest
+      }
+      asset_name=""; url=""; digest="";
+    }'
```
titouanj: This is not working. It should be stored in $res to be used by the lines bellow
And wget isn't writing its result to stdout when used like that. You should do like this:

```suggestion
  res=$(wget --quiet -O- "https://api.github.com/repos/${OWNER}/${REPO}/releases/tags/${RELEASE}" \
| awk -F'"' -v name="$AGENT" -v os="${DISTRIB_VERSION}" -v arch="$ARCH" '
    BEGIN {
    
      version_pattern = "[0-9.-]+"
      pattern_dot = name "\\-" version_pattern "\\." os "\\." arch "\\.rpm$"
      pattern_underscore = name "_" version_pattern "_" arch "_" os"\\.deb$"
    }
    
    # Field detection
    /"name"/ { asset_name = $4 }
    /"browser_download_url"/ { url = $4 }
    /"digest"/ { digest = $4 }
    
    # When all three fields are fullfilled
    (asset_name && url && digest) {
        # Check for pattern
        if (asset_name ~ pattern_dot || asset_name ~ pattern_underscore) {
          print url, digest
      }
      asset_name=""; url=""; digest="";
    }')
```

titouanj (changes requested):

install.sh
```diff
@@ -0,0 +1,406 @@
...
+  tmpdir=$(mktemp -d)
+  http_download "${tmpdir}/${PACKAGE_ID}" "${PACKAGE_URL}"
+  verify_hash "${tmpdir}/${PACKAGE_ID}"
+
+  if test "$LOCAL"; then
+    install_local "${tmpdir}"
+  else
+    case $DISTRIB in
+      ubuntu|debian) sudo apt-get install -yq --allow-downgrades "${tmpdir}/${PACKAGE_ID}" > "/dev/null";;
+      *)  yum install -yq "${tmpdir}/${PACKAGE_ID}" > "/dev/null";;
```
titouanj:
```suggestion
      ubuntu|debian) sudo apt-get install -q --allow-downgrades "${tmpdir}/${PACKAGE_ID}" > "/dev/null";;
      *)  yum install -q "${tmpdir}/${PACKAGE_ID}" > "/dev/null";;
```

install.sh
```diff
@@ -0,0 +1,406 @@
...
+  tmpdir=$(mktemp -d)
+  http_download "${tmpdir}/${PACKAGE_ID}" "${PACKAGE_URL}"
+  verify_hash "${tmpdir}/${PACKAGE_ID}"
+
+  if test "$LOCAL"; then
+    install_local "${tmpdir}"
+  else
+    case $DISTRIB in
+      ubuntu|debian) sudo apt-get install -q --allow-downgrades "${tmpdir}/${PACKAGE_ID}";;
+      *)  yum install -q "${tmpdir}/${PACKAGE_ID}";;
```
titouanj, line 63:
```suggestion
      *)  yum install "${tmpdir}/${PACKAGE_ID}";;
```

titouanj (approved):

ahuret (approved):

PR #386, opened 2026-04-15
test(plugins/cgroups/k8s): Updates to add 80% test code coverage for CGroups/k8s plugin

victoryeagle77:


plugins/cgroups/k8s/tests/k8s.rs
```diff
@@ -64,24 +69,24 @@ fn test_k8s_cgroupv2() -> anyhow::Result<()> {
...
     let cgroup_dir_pod = cgroup_dir_parent.path().join(format!(
         "kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod{POD_UID}.slice"
     ));
-    std::fs::create_dir_all(&cgroup_dir_pod)?;
+    fs::create_dir_all(&cgroup_dir_pod)?;
 
     let source_name = &format!("kubepods-besteffort-pod{POD_UID}");
-    log::info!("cgroup created at {:?}", cgroup_dir_pod);
-    log::info!("source name: {source_name}");
+    info!("cgroup created at {cgroup_dir_pod:?}");
```
TheElectronWill: The project convention is to write `log::info!` explicitely.

plugins/cgroups/k8s/tests/k8s.rs
```diff
@@ -64,24 +69,24 @@ fn test_k8s_cgroupv2() -> anyhow::Result<()> {
 
     // start the measurement pipeline, without the pod's cgroup
     let agent = agent::Builder::new(plugins).build_and_start()?;
-    std::thread::sleep(TOLERANCE);
+    thread::sleep(TOLERANCE);
```
TheElectronWill: Please don't do so many small changes like `mockito::Server` -> `Server` or `std::thread` -> `thread`
1. The change is debatable by itself.
2. When it's in the same PR, and even more so in the same commit, it makes it harder to distinguish what is a "real" change.
3. It "pollutes" the history a little bit.

Testing good practice: in the code that you test, only change what is needed for your tests.

plugins/cgroups/k8s/tests/k8s.rs
```diff
@@ -64,24 +69,24 @@ fn test_k8s_cgroupv2() -> anyhow::Result<()> {
...
-    log::info!("source name: {source_name}");
+    info!("cgroup created at {cgroup_dir_pod:?}");
+    info!("source name: {source_name}");
 
     // expect the source to be created quickly after that
-    std::thread::sleep(TOLERANCE);
+    thread::sleep(TOLERANCE);
     let handle = agent.pipeline.control_handle();
-    let rt = tokio::runtime::Builder::new_current_thread().enable_all().build()?;
+    let rt = runtime::Builder::new_current_thread().enable_all().build()?;
```
TheElectronWill: See, here, I think that it's more clear when prefixed by `tokio`, because the fact that it's tokio is important. But as I said, it's debatable ^^

plugins/cgroups/k8s/src/lib.rs
```diff
@@ -39,7 +42,7 @@ impl AlumetPlugin for K8sPlugin {
         env!("CARGO_PKG_VERSION")
     }
 
-    fn init(config: alumet::plugin::ConfigTable) -> anyhow::Result<Box<Self>> {
+    fn init(config: ConfigTable) -> anyhow::Result<Box<Self>> {
```
TheElectronWill, line 45: I agree that it looks better in trait impl, and it's coherent with the rest of the project. So I'm ok with these changes in `impl AlumetPlugin for K8sPlugin`. But this should be, at least, in a separate commit.
See my other comment.

TheElectronWill (changes requested):

plugins/cgroups/k8s/src/lib.rs
TheElectronWill: it's not in a separate commit :eyes:

plugins/cgroups/k8s/src/lib.rs
```diff
@@ -176,3 +177,26 @@ impl Config {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_default_config() {
```
TheElectronWill: So, basically, you're asserting that `Config::default() == Config::default()`, but with more lines of code?

plugins/cgroups/k8s/src/token.rs
```diff
@@ -231,7 +234,7 @@ impl FromStr for TokenData {
         }
 
         // Decode the payload
-        let payload = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        let payload = general_purpose::URL_SAFE_NO_PAD
```
TheElectronWill: it was fine before IMO

plugins/cgroups/k8s/src/token.rs
```diff
@@ -146,6 +144,11 @@ fn run_kubectl(args: &[&str]) -> Result<String, KubectlError> {
     Ok(content.trim().to_string())
 }
 
+fn run_kubectl(args: &[&str]) -> Result<String, KubectlError> {
+    let output = Command::new("kubectl").args(args).output()?;
```
TheElectronWill, line 149: can you keep the `log::debug!`

plugins/cgroups/k8s/src/token.rs
```diff
@@ -553,4 +568,47 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    fn test_output_kubectl_with_valid_value() {
+        let out = mock_output(ExitStatus::from_raw(0), b"  valid_token \n", b"");
+        let result = output_kubectl(out).unwrap();
+        assert_eq!(result, "valid_token");
```
TheElectronWill: it's a bit confusing because this is _not_ a valid token

"valid string"?

plugins/cgroups/k8s/src/token.rs
```diff
@@ -553,4 +568,47 @@ mod tests {
...
+    #[test]
+    fn test_output_kubectl_with_invalid_utf8() {
+        let out = mock_output(ExitStatus::from_raw(0), &[0xff, 0xfe], b"");
+        let result = output_kubectl(out).unwrap_err();
+        assert!(matches!(result, KubectlError::InvalidOutput));
+    }
+
+    #[test]
+    fn test_output_kubectl_with_status() {
+        // Simulate status code 1
```
TheElectronWill: Why is it not `1` then?

TheElectronWill (changes requested):

plugins/cgroups/k8s/src/token.rs
victoryeagle77: Ah yes you're right it is coming from previous tests

plugins/cgroups/k8s/src/token.rs
victoryeagle77: Okay if you want

plugins/cgroups/k8s/src/token.rs
victoryeagle77: Ah yes it's true I forgot it during the refactor

plugins/cgroups/k8s/src/token.rs
victoryeagle77: It's `1` Unix shifted code but I can render this clearer in writing `1 << 8` if you want

plugins/cgroups/k8s/src/lib.rs
victoryeagle77: Yes for default this test is effectively useless, I can make it more targeted on the values ​​of the fields (and by the same occasion add a quick test for `k8s_node_name`)

plugins/cgroups/k8s/src/token.rs
```diff
@@ -130,10 +131,7 @@ enum KubectlError {
     InvalidOutput,
 }
 
-fn run_kubectl(args: &[&str]) -> Result<String, KubectlError> {
-    log::debug!("running: kubectl {}", args.join(" "));
-    let output = Command::new("kubectl").args(args).output()?;
-
+fn output_kubectl(output: process::Output) -> Result<String, KubectlError> {
```
TheElectronWill:
```suggestion
fn parse_kubectl_output(output: process::Output) -> Result<String, KubectlError> {
```
Let's be nice to our future selves and use super-clear names.

In general, functions are named:
- with a verb first, for actions (like `write()` on File or `poll()` on Source)
- with a name, for getters (for instance, a method that returns the output of a kubectl struct would be named `fn output()`, used like this: `kubectl.output()`

plugins/cgroups/k8s/src/token.rs
TheElectronWill: yes, better! Also, add a comment that explains that it's the _wait status_, not the exist status. (as the doc says https://doc.rust-lang.org/std/os/unix/process/trait.ExitStatusExt.html#tymethod.from_raw)

plugins/cgroups/k8s/src/lib.rs
TheElectronWill: two solutions:
- test with a TOML string (the one from the README, for instance) to have a meaningful test
- don't test and ignore the default_config method (see tarpaulin docs)

TheElectronWill (changes requested):
Almost done

plugins/cgroups/k8s/src/lib.rs
victoryeagle77: Okay, given that we've already tested it with TOML and seen that it didn't do anything, we might as well ignore it.

TheElectronWill (approved):

PR #388, opened 2026-04-20
test(plugins/cgroups/oar): Updates to add 80% test code coverage for CGroups/oar plugin

victoryeagle77:


TheElectronWill (changes requested):
This is not the right approach to the problem.
`util_cgroups::Cgroup` exists for a good reason, and we want to keep it.

To create mock cgroups in your tests, use `CgroupHierarchy::manually_unchecked`.
See `plugins/cgroups/util-cgroups/tests/cgroupv1.rs`

plugins/cgroups/oar/src/attr.rs
```diff
@@ -18,20 +19,22 @@ impl OarJobTagger {
...
         })
     }
-}
 
-impl JobTagger for OarJobTagger {
-    fn attributes_for_cgroup(&mut self, cgroup: &util_cgroups::Cgroup) -> Vec<(String, AttributeValue)> {
-        let extractor = match cgroup.hierarchy().version() {
-            util_cgroups::CgroupVersion::V1 => &self.extractor_v2,
-            util_cgroups::CgroupVersion::V2 => &self.extractor_v3,
+    fn extractor_for_cgroup(&mut self, version: CgroupVersion, path: &str) -> Vec<(String, AttributeValue)> {
```
TheElectronWill: here ^^

plugins/cgroups/oar/src/attr.rs
```diff
@@ -48,3 +51,72 @@ pub fn find_jobid_in_attrs(attrs: &Vec<(String, AttributeValue)>) -> Option<u64>
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_extractor_for_cgroup_with_oar2() {
+        let mut tagger = OarJobTagger::new().unwrap();
+        let attrs = tagger.extractor_for_cgroup(CgroupVersion::V1, "/oar/user_123456");
```
TheElectronWill: and here

plugins/cgroups/oar/src/attr.rs
victoryeagle77: Yes in fact its better like in `job_tracker.rs`

plugins/cgroups/oar/src/config.rs
```diff
@@ -1,36 +0,0 @@
-use std::time::Duration;
```
TheElectronWill, line 1: Why?
This is a PR for tests, don't make non-trivial refactoring (especially without justification).

plugins/cgroups/oar/src/lib.rs
```diff
@@ -119,11 +121,64 @@ impl AlumetPlugin for OarPlugin {
...
     source_setup: source::JobSourceSetup,
     job_cleaner: JobCleaner,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_default_config() {
```
TheElectronWill: I don't believe that this test is worth it. It's basically doing `assert(default_config() == Config::default())`. A test is code, we will need to maintain it in the future. Therefore, we don't add code just to add code (or just to reach the metric we want to).
You have my permission to ignore default_config() ^^

plugins/cgroups/oar/src/source.rs
```diff
@@ -91,29 +110,95 @@ impl CgroupSetupCallback for JobSourceSetup {
     }
 }
 
+fn username_params<F>(id: u64, params: F) -> anyhow::Result<String>
```
TheElectronWill: What is the purpose of this function??

plugins/cgroups/oar/src/source.rs
```diff
@@ -91,29 +110,95 @@ impl CgroupSetupCallback for JobSourceSetup {
...
+    fn test_username_from_id_with_invalid_value() {
+        let result = username_from_id(999_999_999);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_username_params_ok() {
+        let mock = |id| Ok(format!("user_{id}"));
+        let result = username_params(MOCK_ID, mock).unwrap();
+        assert_eq!(result, format!("user_{MOCK_ID}"));
```
TheElectronWill: That's a convoluted way to write
```rust
assert_eq!(format!("user_{MOCK_ID}"), format!("user_{MOCK_ID}"));
```

plugins/cgroups/oar/src/transform.rs
```diff
@@ -30,10 +35,171 @@ impl Transform for JobInfoAttacher {
             if !m.attributes_keys().any(|k| k == "job_id") {
                 // This measurement is not job-specific, attach the list of running jobs.
                 // See issue #209.
-                let jobs_attr = current_job_list.clone();
-                m.add_attr("involved_jobs", AttributeValue::ListU64(jobs_attr));
+                attach_involved_jobs(m, &current_job_list);
```
TheElectronWill: It's not useful to make a function of 2 simple lines and test it independently.
`test_transform` will be just fine

plugins/cgroups/oar/src/attr.rs
```diff
@@ -48,3 +48,85 @@ pub fn find_jobid_in_attrs(attrs: &Vec<(String, AttributeValue)>) -> Option<u64>
...
+            Cgroup::from_cgroup_path(&hierarchy, format!("/oar/user_{}", MOCK_ID[0])), // Tracked job
+            Cgroup::from_cgroup_path(&hierarchy, format!("/oar/user_{}", MOCK_ID[1])), // Not tracked job
+            Cgroup::from_cgroup_path(&hierarchy, "/invalid/job".to_owned()),           // Invalid job
+        ];
+
+        let mut tagger = OarJobTagger::new().unwrap();
+        let attrs = tagger.attributes_for_cgroup(&cgroups[0]);
+
+        assert!(attrs.iter().any(|(k, _)| k == "job_id"));
+        assert!(attrs.iter().any(|(k, _)| k == "user"));
```
TheElectronWill: You should check the value of the attribute, not just the key

plugins/cgroups/oar/src/attr.rs
```diff
@@ -48,3 +48,85 @@ pub fn find_jobid_in_attrs(attrs: &Vec<(String, AttributeValue)>) -> Option<u64>
...
+            Cgroup::from_cgroup_path(&hierarchy, format!("/oar.slice/system/oar-u1-j{}", MOCK_ID[0])), // Tracked job
+            Cgroup::from_cgroup_path(&hierarchy, format!("/oar.slice/system/oar-u1-j{}", MOCK_ID[1])), // Not tracked job
+            Cgroup::from_cgroup_path(&hierarchy, "/invalid/job".to_owned()),                           // Invalid job
+        ];
+
+        let mut tagger = OarJobTagger::new().unwrap();
+        let attrs = tagger.attributes_for_cgroup(&cgroups[0]);
+
+        assert!(attrs.iter().any(|(k, _)| k == "job_id"));
+        assert!(attrs.iter().any(|(k, _)| k == "user_id"));
```
TheElectronWill: same here

TheElectronWill (changes requested):

plugins/cgroups/oar/src/config.rs
victoryeagle77: Because it's a trivial refactoring (just move some lines in an other file), this is related to testing because it's part of the refactoring to centralize tests in one place. Furthermore, this was inconsistent with all the other plugins; none of them had their configuration outside of the lib.rs file.

And as you say after, we will not test the default config, so it is very bad for the plugin coverage to have a file entierly not tested, but if you want I can make this in an other commit.

plugins/cgroups/oar/src/lib.rs
victoryeagle77: Okay for this context it not very usefull

plugins/cgroups/oar/src/source.rs
victoryeagle77: That allows to have a easily testable version of `username_from_id` to increase the coverage

plugins/cgroups/oar/src/source.rs
victoryeagle77: Yes it was a confusion with multiple correction that I made

plugins/cgroups/oar/src/transform.rs
victoryeagle77: OKay, this was the only way I found to cover more than 80% of the file, but it is true that it is not ideal, and thanks to the corrections we will make to the integration test file, we can consider that this will no longer be a problem.

plugins/cgroups/oar/src/attr.rs
victoryeagle77: Yes it's true, I forget to check the values XD, moreover it's not very usefull to check the key.

plugins/cgroups/oar/src/config.rs
TheElectronWill: > […] this was inconsistent with all the other plugins; none of them had their configuration outside of the lib.rs file.

Perhaps, but:
1. It's not something that you change is a "test" PR. A test PR (and test commit) is for tests.
2. There is no real downside of having it in a separate file: it's not a "problem" to fix, so it should stay as it was before.

Regarding the coverage, don't worry :)
1. What matters is the _overall_ coverage of the plugin.
2. If you want, you can write a test that will parse a TOML string and check that the configuration is properly parsed. Not `assert_eq!(serialize_config(deserialize_config(the_config)), the_config))`, but something more interesting.

plugins/cgroups/oar/src/source.rs
TheElectronWill: But in your tests you do:

```rust
let result = username_params(MOCK_USER_ID, |_id| Ok("user".to_string())).unwrap();
assert_eq!(result, "user");
```

which is equivalent to:
```rust
let result = {
  let _ = username_from_id(MOCK_USER_ID); // this is the function that we want to test, and it's not tested
  Ok("user")
};
let result = result.unwrap();
assert_eq!(result, "user");
```

which is actually the same as:
```rust
let result = Ok("user");
assert_eq!(result, Ok("user"));
```
so…

---

And you also do:
```rust
let result = username_params(MOCK_USER_ID, |_id| Err(anyhow!("failed to spawn process id")));
assert!(result.is_err());
```

which is equivalent to:
```rust
assert!(Err(anyhow!("failed to spawn process id")).is_err());
```

IMO, just don't test `username_from_id`.

plugins/cgroups/oar/src/config.rs
victoryeagle77: We had already seen to do something with parsing a TOML string and check that the configuration is properly parsed had no impact on the coverage for default config, but okay I will restore the original file like before.

plugins/cgroups/oar/src/transform.rs
```diff
@@ -37,3 +37,143 @@ impl Transform for JobInfoAttacher {
...
+            buffer
+        };
+
+        let check_output = |ctx: &mut TransformCheckOutputContext| {
+            let measurements = ctx.measurements();
+            let mut iter = measurements.iter();
+            let m = iter.next().expect("measurement expected");
+
+            assert!(iter.next().is_none());
+            assert!(m.attributes_keys().any(|k| k == "job_id"));
```
TheElectronWill: Check the value of the job id

plugins/cgroups/oar/src/transform.rs
```diff
@@ -37,3 +37,143 @@ impl Transform for JobInfoAttacher {
...
+            assert!(!m.attributes_keys().any(|k| k == "involved_jobs"));
```
TheElectronWill, line 166: Why is there no involved jobs if you added them to the job tracker earlier?

TheElectronWill (changes requested):

plugins/cgroups/oar/src/config.rs
TheElectronWill: Coverage % is not the main objective, it's just a metric.
The true goal is to ensure that the code works _in the long term_ :wink:

plugins/cgroups/oar/src/transform.rs
victoryeagle77: `involved_jobs` is only attached to measurements that don't already containing a `job_id`.

plugins/cgroups/oar/src/transform.rs
victoryeagle77: Yes I made the modification in an other file with a similar test but I forget to make the same thing here

plugins/cgroups/oar/src/source.rs
```diff
@@ -91,29 +110,102 @@ impl CgroupSetupCallback for JobSourceSetup {
...
-    let output = child
-        .wait_with_output()
-        .context("failed to wait for process id to terminate")?;
-    if !output.status.success() {
-        let error_message = String::from_utf8_lossy(&output.stderr).into_owned();
-        return Err(anyhow!("process id failed with {}", output.status).context(error_message));
-    }
-    let username = String::from_utf8_lossy(&output.stdout).into_owned();
-    Ok(username)
+    username_params(id, |id| {
```
TheElectronWill: please revert this, see https://github.com/alumet-dev/alumet/pull/388#discussion_r3303585166
(It seems that I cannot mark one occurrence of this remark as a duplicate without marking it everywhere… thanks github)

TheElectronWill (changes requested):

plugins/cgroups/oar/tests/oar.rs
```diff
@@ -14,174 +12,175 @@ use alumet::{
...
-    assert!(elements.is_empty(), "source should have been deleted: {source_name}");
-
-    // stop the pipeline and wait for it to terminate
-    handle.shutdown();
-    agent.wait_for_shutdown(TIMEOUT)?;
-    Ok(())
-}
+// #[test]
+// #[serial]
+// fn test_oar3() -> anyhow::Result<()> {
```
TheElectronWill: if you want to disable this, don't comment the code, add SKIP_CGROUPFS_TEST=true to your env (it will change in the future, see the related task in the sprint)

TheElectronWill (changes requested):

plugins/cgroups/oar/tests/oar.rs
victoryeagle77: No I only forget to remove the comment to disable before push LOL 
I will do that in an other PR because we need to apply this for every modules of cgroups plugin

TheElectronWill (approved):

PR #389, opened 2026-04-21
test(plugins/cgroups/util-cgroups-plugins): Updates to add 80% test code coverage for CGroups/util-cgroups-plugins plugin

victoryeagle77:


plugins/cgroups/util-cgroups-plugins/src/cpus.rs
```diff
@@ -32,8 +32,43 @@ fn parse_cpu_list(cpulist: &str) -> anyhow::Result<Vec<u32>> {
     Ok(cpus)
 }
 
-pub fn online_cpus() -> anyhow::Result<Vec<u32>> {
-    let path = "/sys/devices/system/cpu/online";
-    let list = std::fs::read_to_string(path).with_context(|| format!("failed to parse {path}"))?;
+fn online_cpus_path(path: &std::path::Path) -> anyhow::Result<Vec<u32>> {
```
TheElectronWill, line 35: Is it not enough to test `parse_cpu_list`? That would be much simpler.

plugins/cgroups/util-cgroups-plugins/src/job_annotation_transform.rs
```diff
@@ -190,4 +192,45 @@ mod tests {
...
         thread_b.join().unwrap();
     }
+
+    #[test]
+    fn test_on_cgroupfs_mounted_sets_v2() {
+        let shared = SharedCgroupHierarchy::default();
+        let mut optional = OptionalSharedHierarchy::default();
+        optional.enable(shared.clone());
+
+        let hierarchies = vec![
```
TheElectronWill, line 202: the two hierarchies should have separate paths, and you should check that the path is correct at the end

plugins/cgroups/util-cgroups-plugins/src/self_stop.rs
```diff
@@ -17,3 +17,33 @@ pub fn analyze_io_result<R>(res: io::Result<R>) -> Result<R, PollError> {
...
+        let path = tmp.path().join("test_file");
+        write(&path, "content").unwrap();
+        let result = analyze_io_result(read_to_string(&path));
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_analyze_io_result_invalid() {
+        let tmp = tempfile::tempdir().unwrap();
+        let path = tmp.path().join("invalid");
```
TheElectronWill:
```suggestion
        let path = tmp.path().join("i_do_not_exist");
```

plugins/cgroups/util-cgroups-plugins/tests/tests_metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+};
+use util_cgroups_plugins::metrics::AugmentedMetrics;
+
+const PLUGIN_NAME: &str = "test-util-cgroups-plugin-metrics";
+
+struct MockPlugin;
+
+impl AlumetPlugin for MockPlugin {
+    fn name() -> &'static str {
+        PLUGIN_NAME
```
TheElectronWill, line 23: you can inline this constant, it's used only once

TheElectronWill (changes requested):

plugins/cgroups/util-cgroups-plugins/src/metrics.rs
```diff
@@ -156,18 +156,3 @@ impl AugmentedMetrics {
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn augmented_metrics() {
```
TheElectronWill, line 165: ?

plugins/cgroups/util-cgroups-plugins/src/cpus.rs
victoryeagle77: Yes but if we do that, we loose tests for the file and content reading, and knowing that I did the same thing for RAPL, might as well do the same here

plugins/cgroups/util-cgroups-plugins/src/metrics.rs
victoryeagle77: Yes, this test is useless now with `test_metrics` file but if it was only in the scope to check a compiling behavior, I can restore it.

plugins/cgroups/util-cgroups-plugins/src/cpus.rs
TheElectronWill: > we loose tests for the file and content reading

"and nothing of value was lost"

I'm not 100% convinced that it brings a lot of value to the tests (`std::fs::read_to_string` is not our responsibility and is super simple), but okay, let's keep it.

TheElectronWill (changes requested):
One last thing

plugins/cgroups/util-cgroups-plugins/tests/mod.rs
```diff
@@ -0,0 +1,2 @@
+#[cfg(test)]
+pub mod tests_metrics;
```
TheElectronWill: Remove this file, it's not needed here.

https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory
> Each file in the tests directory is a separate crate, so we need to bring our library into each test crate’s scope.

TheElectronWill: You only need this `mod.rs` and module declaration when you are creating modules in `src/`. Integration tests, in `tests/`, are a bit different.

victoryeagle77: Yes it's useless now I forgot to remove it during refactor that I've made in src/ before that.

TheElectronWill (approved):

PR #392, opened 2026-04-28
Feat/brseed 1457 nvml new metric

Fyspeed:
Add a new necessary metrics: GPU VRAM allocation (called nvml_gpu_memory_allocation)

Similar to Procfs 'memory_usage' field, this metric measures the allocation of the VRAM. It has an attribute "kind":

- free : the amout of bytes not allocated
- total: the total size of the VRAM
- used: the amount of bytes currently used
- reserved: the amount of bytes reserved for sytem use (may not be available depending on GPU model / NVIDIA driver version)

Also updates the README to add this new metrics

### DIFF
--- a/core/alumet/src/test/runtime.rs
+++ b/core/alumet/src/test/runtime.rs
@@ -1,6 +1,5 @@
 use std::{
     cell::RefCell,
-    collections::HashMap,
     ops::Deref,
     rc::Rc,
     sync::{Arc, Mutex},
@@ -16,7 +15,7 @@ use wrapped_transform::{SetTransformOutputCheck, TransformDone, WrappedTransform
 
 use crate::{
     agent::builder::TestExpectations,
-    measurement::{MeasurementBuffer, MeasurementPoint, MeasurementType},
+    measurement::{AttributeValue, MeasurementBuffer, MeasurementPoint, MeasurementType},
     metrics::{
         Metric, RawMetricId,
         duplicate::{DuplicateCriteria, DuplicateReaction},
@@ -712,7 +711,9 @@ impl<'a> SourceCheckOutputContext<'a> {
         self.metrics
     }
 
-    pub fn points_by_metric_and_consumer(&'a self) -> IndexMap<(&'a str, ResourceConsumer), &'a MeasurementPoint> {
+    pub fn points_by_metric_and_consumer(
+        &'a self,
+    ) -> IndexMap<(&'a str, ResourceConsumer, Option<AttributeValue>), &'a MeasurementPoint> {
         self.measurements()
             .into_iter()
             .map(|p| {
@@ -721,8 +722,19 @@ impl<'a> SourceCheckOutputContext<'a> {
                     .by_id(&p.metric)
                     .unwrap_or_else(|| panic!("unknown metric with id {} in {p:?}", &p.metric.0));
                 let metric_name = metric.name.as_str();
+
+                // If the metric has an attribute "kind", the value of the attribute is stored
+                // Otherwise, the Option<AttributeValue> field should be equal to None
+                let kind = p
+                    .attributes()
+                    .find(|(key, _value)| match *key {
+                        "kind" => true,
+                        _ => false,
+                    })
+                    .map(|(_key, value)| value.clone());
+
                 let consumer = p.consumer.clone();
-                ((metric_name, consumer), p)
+                ((metric_name, consumer, kind), p)
             })
             .collect()
     }
--- a/core/alumet/src/test/runtime.rs
+++ b/core/alumet/src/test/runtime.rs
@@ -713,6 +713,7 @@ impl<'a> SourceCheckOutputContext<'a> {
 
     pub fn points_by_metric_and_consumer(
         &'a self,
+        attribute_key: Option<&str>,
     ) -> IndexMap<(&'a str, ResourceConsumer, Option<AttributeValue>), &'a MeasurementPoint> {
         self.measurements()
             .into_iter()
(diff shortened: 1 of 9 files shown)


core/alumet/src/test/runtime.rs
```diff
@@ -721,8 +722,19 @@ impl<'a> SourceCheckOutputContext<'a> {
                     .by_id(&p.metric)
                     .unwrap_or_else(|| panic!("unknown metric with id {} in {p:?}", &p.metric.0));
                 let metric_name = metric.name.as_str();
+
+                // If the metric has an attribute "kind", the value of the attribute is stored
```
TheElectronWill, line 728: This method is in alumet core, it should not depend on things that are too specific.
You can add an "attribute" parameter to choose the attribute, or make another version of the function in your tests (not in this file).

plugins/nvidia-nvml/src/probe.rs
```diff
@@ -140,6 +140,54 @@ impl<D: NvmlDevice> Source for FullSource<D> {
...
 
+        // Get the current memory allocation information for this device in Bytes
+        if features.memory_allocation {
+            measurements.push(
+                MeasurementPoint::new(
+                    timestamp,
+                    self.metrics.memory_allocation,
+                    self.resource.clone(),
+                    consumer.clone(),
+                    device.memory_allocation()?.free,
```
TheElectronWill: You should call `device.memory_allocation()?` only once, and then look at its fields.

plugins/nvidia-nvml/src/nvml/mod.rs
```diff
@@ -47,6 +47,10 @@ pub trait NvmlDevice: Display + Send {
     /// See [`nvml_wrapper::Device::utilization_rates`].
     fn utilization_rates(&self) -> NvmlResult<nvml_wrapper::struct_wrappers::device::Utilization>;
 
+    /// Current memory usage.
+    /// See [`nvml_wrapper::Device::memory_allocation`].
+    fn memory_allocation(&self) -> NvmlResult<nvml_wrapper::struct_wrappers::device::MemoryInfo>;
```
TheElectronWill: To make things easier, I think it's better to keep the original name `memory_info`

plugins/nvidia-nvml/src/probe.rs
```diff
@@ -140,6 +140,54 @@ impl<D: NvmlDevice> Source for FullSource<D> {
...
+                )
+                .with_attr("kind", "free"),
+            );
+            // According to NVIDIA documentation, system-reserved memory field
+            // only exists from version 2 onward
+            if device.memory_allocation()?.version > 1 {
```
TheElectronWill: This is always the case when calling `nvmlDeviceGetMemoryInfo_v2`. The previous version (`nvmlDeviceGetMemoryInfo`) returns a different structure with different fields.

Also, how do we call the previous version of `nvmlDeviceGetMemoryInfo_v2` does not exist? (it seems that this is not supported by nvml-wrapper at the moment… sad)

TheElectronWill (changes requested):

plugins/nvidia-nvml/README.md
```diff
@@ -30,6 +31,22 @@ One source will be created per GPU device.
 
 Some metrics can be disabled, see the `mode` configuration option.
 
+### Attributes
+
+#### Kind
+
+The kind of the memory is the type of the allocated memory space reserved by the system or the hardware.
+Depending on the the version of the version of your NVIDIA drivers and your GPU model, the output may slightly vary. See:
```
TheElectronWill: In the readme, we could be a bit higher level (easier for the end user).
I would talk in terms of driver version, not NVML struct version. According to the [changelog](https://docs.nvidia.com/deploy/nvml-api/change-log.html#change-log), the v2 appeared in v510.

plugins/nvidia-nvml/README.md
```diff
@@ -74,3 +91,11 @@ For instance, to obtain non-zero values for the video encoding/decoding metrics,
 NVML requires 20-100ms to refresh counter values based on GPU model.
 When `poll_interval` is set too low, the plugin queries identical counter values repeatedly during polling.
 Since some measurements are calculated from previous polls, these measurements are discarded rather than reported as zero values.
+
+## Note on NVIDIA 'utilization'
```
TheElectronWill, line 92: good idea to explain that

TheElectronWill:
One more thing: instead of adding commits to fix the linter, do `git rebase -i` and amend your previous commits. Let's keep the history clean :)

plugins/nvidia-nvml/src/probe.rs
Fyspeed: I've only seen one function in the nvml-wrapper named["memory info"](https://docs.rs/nvml-wrapper/0.12.0/nvml_wrapper/device/struct.Device.html#method.memory_info)
However the struct it returns is different depending on the version (but it has the same name)

Also older versions of the struct don't have a field "version", unlike recent versions ([example](https://docs.rs/nvml-wrapper/0.5.0/nvml_wrapper/struct_wrappers/device/struct.MemoryInfo.html)

So I guess a way to find out which version is currently used would be to access the field "version", and if it's unitialized/throws an error, we know it's an older version of the struct ?

Fyspeed: Or is it possible to access the version of NVML in the code ? Then i can just check if it's < v510 or >= v510

edit: found this in the doc https://docs.rs/nvml-wrapper/0.12.0/nvml_wrapper/struct.Nvml.html#method.sys_nvml_version, so i'm gonna try to use that

plugins/nvidia-nvml/src/lib.rs
```diff
@@ -469,64 +479,108 @@ mod tests {
                     // first trigger
                     let points = out.points_by_metric_and_consumer();
 
-                    assert_eq!(points.len(), 10, "wrong number of points, got {points:?}");
+                    assert_eq!(points.len(), 14, "wrong number of points, got {points:?}");
 
                     assert_eq!(
-                        points[&("nvml_encoder_utilization", ResourceConsumer::LocalMachine)]
+                        points[&("nvml_encoder_utilization", ResourceConsumer::LocalMachine, None)]
```
AngeCyp, line 485: Not a huge fan of this (see related comment)

AngeCyp (changes requested):

plugins/nvidia-nvml/src/lib.rs
Fyspeed: The key has to include the attribute value in some way; how would you change that ?

i've updated the function points_by_metric_and_consumer to only retrieve chosen attributes

TheElectronWill (approved):

AngeCyp (approved):

PR #399, opened 2026-05-06
test(plugins/cgroups/util-cgroups): Updates to add 80% test code coverage for CGroups/util-cgroups plugin

victoryeagle77:


plugins/cgroups/util-cgroups/benches/parse.rs
```diff
@@ -412,9 +372,43 @@ mod tests {
...
+        assert!(!b128.contains(4));
+        assert!(b128.contains(5));
+        assert!(!b128.contains(6));
+        assert!(b128.contains(7));
+    }
+
     #[test]
     fn test_parse_space_kv_unchecked_cached_indices_hashset() {
-        let data = b"a1 111\nb2 222\nc3 333\n";
+        let data = b"key1 111\nkey2 222\nkey3 333\n";
```
TheElectronWill, line 411: How does this change improve the test?

plugins/cgroups/util-cgroups/src/measure/bitset.rs
```diff
@@ -80,4 +80,12 @@ mod tests {
         assert!(set.contains(12));
         assert_eq!(set, BitSet128::new(&[0, 1, 2, 12, 25, 32, 64, 111, 127]))
     }
+
+    #[test]
+    fn test_fmt() {
+        let set = format!("{:?}", BitSet128::new(&[0, 2, 3]));
+        assert!(set.contains("BitSet128"));
+        // 1<<0 + 1<<2 + 1<<3 = 13(10) = 1101(2)
+        assert!(set.contains("1101"));
```
TheElectronWill:
```suggestion
        assert_eq!(set, "BitSet128(1101)");
```
or something like that

plugins/cgroups/util-cgroups/src/mount_wait.rs
```diff
@@ -73,34 +73,42 @@ impl Drop for CgroupMountWait {
     }
 }
 
+fn handle_event(
```
TheElectronWill, line 76: This extraction is a good idea :)

plugins/cgroups/util-cgroups/tests/cgroupv1.rs
```diff
@@ -68,7 +68,6 @@ fn one_cgroup_created_v1() {
 
 #[test]
 fn one_cgroup_created_after_v1() {
-    env_logger::init_from_env(env_logger::Env::default());
```
TheElectronWill, line 71: ?

TheElectronWill (changes requested):

plugins/cgroups/util-cgroups/tests/cgroupv1.rs
victoryeagle77: Yes, the previous line before the modification causes flacky errors due to its initialization (occurs sometimes only for tarpaulin, sometimes only for cargo test). But in fact the env_logger here is globally not very usefull.

plugins/cgroups/util-cgroups/benches/parse.rs
victoryeagle77: Because before we had a mix of letters and "key" symbols between tests, this is more uniform and cleaner, it's simply more aesthetically.

plugins/cgroups/util-cgroups/src/measure/bitset.rs
victoryeagle77: Yes you're right it's the same and more clearer like that.

plugins/cgroups/util-cgroups/tests/cgroupv1.rs
TheElectronWill: Ok, that can probably be solved by using [try_init](https://docs.rs/env_logger/latest/env_logger/fn.try_init.html) or `try_init_from_env` instead of `init_from_env`.
But let's remove it this time.

TheElectronWill (approved):

PR #400, opened 2026-05-06
feat(plugins/opentelemetry): Replace PeriodicReader with direct gRPC export

ValentinCB:
# Description
A output plugin which exports Alumet metrics to an OpenTelemetry collector via gRPC.

This PR replaces the previous `opentelemetry-sdk`-based implementation that used a `PeriodicReader`, which capped the export frequency at 1Hz regardless of Alumet's pipeline cadence. The new implementation calls the OTLP gRPC endpoint directly inside `write()`, allowing metrics to be exported at any frequency.

Functional tests are also added to cover the main export behaviors.

For more info, please refer to `./plugins/opentelemetry/README.md`.

# How to test
At the root of the repo, edit the `alumet-config.toml` file. **Lower the pool intervals of RAPL and Procfs under 1s**, and add the `[opentelemetry]` config field:

```toml
[plugins.opentelemetry]
# URL of the OTLP gRPC collector
collector_host = "http://localhost:4317"
# Optional prefix and suffix for metric names
prefix = ""
suffix = ""
# Use the display name for units (e.g. "gCO₂") instead of the unique name (e.g. "g_CO2")
use_unit_display_name = false
# Forward measurement attributes as OTel data-point attributes
add_attributes_to_labels = false
```

Start a local OpenTelemetry collector (e.g. with Docker):

```bash
docker run -p 4317:4317 otel/opentelemetry-collector
```

Build the binary and launch Alumet with `rapl` and `opentelemetry`:

```bash
cargo build
./target/debug/alumet-agent --config alumet-config.toml
```

You should see metrics being exported to your collector in real time.

To run the functional tests:

```bash
cargo test --test output -p plugin-opentelemetry
```


plugins/opentelemetry/src/lib.rs
```diff
@@ -47,13 +46,12 @@ impl AlumetPlugin for OpenTelemetryPlugin {
...
 #[derive(Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
-struct Config {
-    collector_host: String,
-    prefix: String,
-    suffix: String,
-    use_unit_display_name: bool,
-    add_attributes_to_labels: bool,
-    push_interval_seconds: u64,
+pub struct Config {
```
guigomcha, line 49: does it really need to be pub?

plugins/opentelemetry/src/output.rs
```diff
@@ -76,69 +63,141 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
+            let value_as_f64 = match m.value {
+                WrappedMeasurementValue::F64(v) => v,
+                WrappedMeasurementValue::U64(v) => v as f64,
+            };
+
+            let time_unix_nano = m
+                .timestamp
+                .duration_since(UNIX_EPOCH.into())
+                .map(|d| d.as_nanos() as u64)
+                .unwrap_or(now_ns);
```
guigomcha: is it possible to have a metric in the buffer that does not have a timestamp? Given that we have the flush vs poll interval differences, If this happens I think that we should simply fail tbh

plugins/opentelemetry/src/output.rs
```diff
@@ -76,69 +63,141 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
+
+            // Prepare the Data Point
+            let value_as_f64 = match m.value {
+                WrappedMeasurementValue::F64(v) => v,
+                WrappedMeasurementValue::U64(v) => v as f64,
+            };
+
+            let time_unix_nano = m
+                .timestamp
+                .duration_since(UNIX_EPOCH.into())
```
guigomcha: Not sure how the timestamp works in the measurement from alumet, but why the delta instead of the total timestamp?

plugins/opentelemetry/tests/fakeplugin.rs
guigomcha, line 1: There is no assert or creation of the output plugin in this test right? it'd be nice to ensure that when the buffer has 2 data points of the same metric, then the otput generates the corresponding entries in the gauge's data points

guigomcha (changes requested):
Super nice, just a few questions

plugins/opentelemetry/src/lib.rs
ValentinCB: I need it to be public for testing :

```
error[E0603]: struct `Config` is private
  --> plugins/opentelemetry/tests/output.rs:21:28
   |
21 | use plugin_opentelemetry::{Config, OpenTelemetryPlugin};
   |                            ^^^^^^ private struct
   |
note: the struct `Config` is defined here
  --> /home/circleci/project/plugins/opentelemetry/src/lib.rs:49:1
   |
49 | struct Config {
   | ^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0603`.
error: could not compile `plugin-opentelemetry` (test "output") due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

plugins/opentelemetry/src/output.rs
ValentinCB: `m.timestamp` is an absolute timestamp already (Alumet wraps a SystemTime internally). The `duration_since(UNIX_EPOCH)` call was only converting it into the Unix epoch format expected by OTLP, not computing a delta between measurements. 

`duration_since(UNIX_EPOCH)` does compute a delta, but specifically the delta between the timestamp and the Unix epoch (1970-01-01), which is equivalent to convert absolute timestamp to UNIX timestamp.

But Alumet already provide a method to convert absolute timestamp to UNIX so I modified the code to use it instead:
```
let (secs, nanos) = m.timestamp.to_unix_timestamp();
let time_unix_nano = secs * 1_000_000_000 + nanos as u64;
```

plugins/opentelemetry/tests/fakeplugin.rs
ValentinCB: This is actually what `group_multiple_datapoints_under_same_metric`  is intended to verify.

The test checks that:

- only one OTLP Metric is generated for "dummy",
- and that its Gauge contains two data_points.

guigomcha: I checked it before we had the tests/output.rs. All good now

guigomcha (approved):
LGTM! nice improvement!

plugins/opentelemetry/src/output.rs
```diff
@@ -76,65 +60,125 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
+                    m.consumer.id_string().unwrap_or("empty".to_string()),
                 ),
             ];
+
             if self.add_attributes_to_labels {
-                // Add attributes as labels
                 for (key, value) in m.attributes() {
-                    labels.push(KeyValue::new(key.to_owned(), value.to_string()));
+                    let v = value.to_string();
+                    let v = if v.is_empty() { "empty".to_string() } else { v };
```
ahuret, line 92: this behavior should be documented in the readme

plugins/opentelemetry/src/output.rs
```diff
@@ -76,65 +60,125 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
+            attributes.sort_by(|a, b| a.key.cmp(&b.key));
+
+            // Prepare the Data Point
+            let value_as_f64 = match m.value {
+                WrappedMeasurementValue::F64(v) => v,
+                WrappedMeasurementValue::U64(v) => v as f64,
             };
+
+            let (secs, nanos) = m.timestamp.to_unix_timestamp();
+            let time_unix_nano = secs * 1_000_000_000 + nanos as u64;
```
ahuret:
```suggestion
            let time_unix_nano = m.timestamp.duration_since(std::time::UNIX_EPOCH.into())
                .map_err(|e| anyhow::anyhow!("invalid timestamp: {e}"))?
                .as_nanos() as u64;
```

a more idiomatic way

plugins/opentelemetry/src/output.rs
```diff
@@ -76,65 +60,125 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
+
+            let data_point = NumberDataPoint {
+                attributes,
+                start_time_unix_nano: 0,
+                time_unix_nano,
+                value: Some(opentelemetry_proto::tonic::metrics::v1::number_data_point::Value::AsDouble(value_as_f64)),
```
ahuret: add `use opentelemetry_proto::tonic::metrics::v1::number_data_point::Value` at the top of the file and use it here

plugins/opentelemetry/README.md
```diff
@@ -2,11 +2,13 @@
...
 
-Implements a push-based exporter (via gRPC) which can be connected to an OpenTelemetry Collector (via a receiver), processed in any way, and then exported to a observability backend like Jaeger, Prometheus, Thanos, OpenSearch, ElasticSearch, etc.
+Implements a push-based exporter (via gRPC) which can be connected to an OpenTelemetry Collector (via a receiver), processed in any way, and then exported to an observability backend like Jaeger, Prometheus, Thanos, OpenSearch, ElasticSearch, etc.
+
+Each call to `write()` triggers an immediate gRPC export, allowing metrics to be exported at any frequency.
 
 ## Requirements
 
-- It needs to have a OTEL Collector reachable.
+- A reachable OpenTelemetry Collector with a gRPC receiver enabled (default port: `4317`).
```
ahuret:
```suggestion
- A reachable OpenTelemetry Collector with an OTLP/gRPC receiver enabled.
```

plugins/opentelemetry/README.md
```diff
@@ -2,11 +2,13 @@
 
 This crate is a library that defines the OpenTelemetry plugin.
 
-Implements a push-based exporter (via gRPC) which can be connected to an OpenTelemetry Collector (via a receiver), processed in any way, and then exported to a observability backend like Jaeger, Prometheus, Thanos, OpenSearch, ElasticSearch, etc.
+Implements a push-based exporter (via gRPC) which can be connected to an OpenTelemetry Collector (via a receiver), processed in any way, and then exported to an observability backend like Jaeger, Prometheus, Thanos, OpenSearch, ElasticSearch, etc.
```
ahuret:
```suggestion
Implements a push-based exporter (via OTLP/gRPC) which can be connected to an OpenTelemetry Collector (via a receiver), processed in any way, and then exported to an observability backend like Jaeger, Prometheus, Thanos, OpenSearch, ElasticSearch, etc.
```

ahuret (changes requested):

plugins/opentelemetry/README.md
```diff
@@ -15,15 +17,16 @@ Put the following in the configuration file of the Alumet agent (usually `alumet
...
 prefix = ""
 suffix = "_alumet"
 # Use the display name of the units instead of their unique name, as specified by the UCUM.
-# See https://ucum.org/ucum for a list of unit and their symbols.
+# See https://ucum.org/ucum for a list of units and their symbols.
+# Example: "W" (display) vs "watt" (unique)
 use_unit_display_name = true
+# Forward measurement attributes as OpenTelemetry data-point attributes
 add_attributes_to_labels = true
 ```
```
ahuret, line 32: It would be nice to add a section in "More Information" that explain how frequently the requests will be sent to the otel collector, explaining that it depends on the flush_interval of sources  (but explaining that it's decoupled from the poll_interval concept etc)

plugins/opentelemetry/README.md
```diff
@@ -15,18 +17,95 @@ Put the following in the configuration file of the Alumet agent (usually `alumet
...
 
 ## More information
 
 Check more at the [user-book website](https://alumet-dev.github.io/user-book/plugins/output/opentelemetry.html).
+
+> **Note**: If a point with an empty value is received, this plugin will set it's value to `"empty"` before forwarding it to OpenTelemetry.
```
ahuret:
```suggestion
> **Note**: If a point with an empty attribute value is received, this plugin will set it's value to `"empty"` before forwarding it to OpenTelemetry.
```

plugins/opentelemetry/README.md
```diff
@@ -15,18 +17,95 @@ Put the following in the configuration file of the Alumet agent (usually `alumet
...
+flush_interval = "4s"
+# Which method to use to obtain memory statistics.
+memory_mode = "quick"
+
+# ...
+```
+
+RAPL continues to flush 2400 points every 2s as before. **In parallel**, procfs flushes (4s / 0.500s) × 9 = **72 points every 4s**.
+
+Each source flushes independently: the Alumet/OTEL plugin receives a separate `MeasurementBuffer` from each and issues a dedicated OTLP/gRPC request for each — concurrently. The resulting traffic looks like:
```
ahuret: this is actually not totally true, as procfs will create one source per process, the number of batchs at the end depends on the number process watched by alumet (in a simple case with one process using alumet exec there's a single source but in usual watch mode you would have lot of them in //).

On its side, RAPL plugin always have only one source

ValentinCB: got it, I'll update the README.md

ahuret (dismissed):

plugins/opentelemetry/tests/output.rs
```diff
@@ -0,0 +1,797 @@
...
+    measurement::{MeasurementBuffer, MeasurementPoint, Timestamp, WrappedMeasurementValue},
+    pipeline::naming::OutputName,
+    plugin::PluginMetadata,
+    resources::{Resource, ResourceConsumer},
+    test::{RuntimeExpectations, runtime::OutputCheckInputContext},
+};
+
+/// Serialise all tests to avoid races on opentelemetry's process-wide global
+/// meter provider (`global::set_meter_provider`).
+static TEST_MUTEX: Mutex<()> = Mutex::new(());
```
TheElectronWill: I agree with the principle.
For the implementation, you can use the [serial_test](https://docs.rs/serial_test/latest/serial_test/) crate, it will be easier to read and it will handle edge cases better.

plugins/opentelemetry/src/output.rs
```diff
@@ -76,65 +62,128 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
-            self.initialize();
-            self.initialized = true;
-        }
-        let common_scope_attributes = vec![KeyValue::new("tool", "alumet")];
-        let scope = InstrumentationScope::builder("alumet")
-            .with_version(env!("CARGO_PKG_VERSION"))
-            .with_attributes(common_scope_attributes)
-            .build();
+
+        // Map where the key is the full metric name (including prefix/suffix)
```
TheElectronWill: This comment deserves to be on the field itself, like
```rust
struct OpenTelemetryOutput {
  …
  ///  Map where the key is the full metric name (including prefix/suffix)
  metric_map
}
```

plugins/opentelemetry/src/output.rs
```diff
@@ -76,65 +62,128 @@ impl alumet::pipeline::Output for OpenTelemetryOutput {
...
-                .with_unit(get_unit_string(full_metric, self.use_unit_display_name))
-                .build();
-            match m.value {
-                WrappedMeasurementValue::F64(v) => gauge.record(v as f64, &labels),
-                WrappedMeasurementValue::U64(v) => gauge.record(v as f64, &labels),
+
+            attributes.sort_by(|a, b| a.key.cmp(&b.key));
+
+            // Prepare the Data Point
+            let value_as_f64 = match m.value {
```
TheElectronWill: IIRC you can do `m.value.as_f64()` :)

TheElectronWill (changes requested):

TheElectronWill (approved):

PR #406, opened 2026-06-03
test(plugins/cgroups): Invert how the manual integration tests work in cgroup plugins

victoryeagle77:


TheElectronWill (commented):
the CI needs to be updated, it currently uses SKIP_CGROUPFS_TESTS (that can be removed)

.circleci/config.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           no_output_timeout: 30s
           environment:
             # tests that manipulate the cgroupfs don't work in the CI container, because it has no systemd user slice, so the tests cannot create cgroups
-            SKIP_CGROUPFS_TESTS: true
+            RUN_CGROUPFS_TESTS: true
```
TheElectronWill: cgroupfs tests must not run in the CI…

TheElectronWill (changes requested):

.circleci/config.yml
TheElectronWill: this is not resolved :p

plugins/cgroups/k8s/tests/k8s.rs
```diff
@@ -25,8 +25,11 @@ const TOKEN_CONTENT: &str = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMj
 
 #[test]
 fn test_k8s_cgroupv2() -> anyhow::Result<()> {
-    if std::env::var_os("SKIP_CGROUPFS_TESTS").is_some() {
-        println!("skipped because SKIP_CGROUPFS_TESTS is set");
+    if !matches!(
+        std::env::var("RUN_CGROUPFS_TESTS").as_deref(),
+        Ok("true") | Ok("yes") | Ok("1")
```
titouanj: Why matching multiple values?

victoryeagle77: It's an idea of @TheElectronWill, to be modular with the value that we want to give to `RUN_CGROUPFS_TESTS`

titouanj: Ok why not. But to be sure that we won't break something about this in the future, shouldn't we factorize this in a small common function?

victoryeagle77: I don't think it's very usefull, knowing those conditions will be removed in future, and replaced by a mock of  the cgroup path, to realy run the tests in CI.

TheElectronWill: why didn't you use a `if let Ok("true" | "yes" | "1")` like I suggested?

TheElectronWill: edit: regarding the function, it's a good idea. If it requires too much work we can postpone it, or do it just where it's simple to do.

TheElectronWill (changes requested):

plugins/cgroups/util-cgroups/src/lib.rs
```diff
@@ -12,3 +12,14 @@ pub mod mount_wait;
 pub use detect::CgroupDetector;
 pub use hierarchy::{Cgroup, CgroupHierarchy, CgroupVersion};
 pub use mount_wait::CgroupMountWait;
+
+pub fn run_cgroups_tests() -> anyhow::Result<()> {
```
TheElectronWill: That should not be in the public API of this crate, because it's only used in tests.
This is why I suggested to postpone it.

Also, you still have a complicated `matches!`

TheElectronWill (changes requested):

plugins/cgroups/util-cgroups-plugins/tests/tests.rs
```diff
@@ -409,11 +401,9 @@ fn test_no_cgroupv2_at_all() -> anyhow::Result<()> {
 }
 
 #[test]
-fn test_on_cgroupfs_mounted() -> () {
-    if std::env::var_os("SKIP_CGROUPFS_TESTS").is_some() {
-        println!("skipped because SKIP_CGROUPFS_TESTS is set");
-        return ();
-    }
+fn test_on_cgroupfs_mounted() -> anyhow::Result<()> {
+    run_cgroups_tests()?;
```
TheElectronWill: IMO it's confusing to use it that way, especially with that name.
```
if !run_cgroups_tests() {
  return …
}
```
was good.

TheElectronWill (approved):

PR #408, opened 2026-06-05
test(core): fix flaky test control::source_flush

TheElectronWill:


core/alumet/tests/control.rs
```diff
@@ -273,17 +273,14 @@ fn source_flush() {
...
     let n_flushed_after_flush = counters.n_transform_in.load(COUNTER_ORD);
     assert_ne!(n_flushed_after_flush, 0);
 
     // check that the manual flush happens only once
     println!("waiting more");
     std::thread::sleep(Duration::from_millis(500));
     println!("checking not flushed");
-    let n_written_later = counters.n_written.load(COUNTER_ORD);
     let n_flushed_later = counters.n_transform_in.load(COUNTER_ORD);
-    assert!(n_written_later > n_written_after_flush);
```
titouanj, line 286: Why was this assertion failing randomly?

titouanj (approved):

core/alumet/tests/control.rs
TheElectronWill: My best explanation is that it's the combo sleep + assertion on n_written: it depended on the latency of the full pipeline source->transform->output

Since it was irrelevant to what I actually wanted to test (source->), I figured that I could just remove it.

PR #409, opened 2026-06-08
docs: rewrite the contribution guidelines

TheElectronWill:
[Rendered](https://github.com/alumet-dev/alumet/blob/docs/contributing-ux/CONTRIBUTING.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,89 +1,259 @@
-# Contributing guide
+# Contribution Guidelines
 
-Hello and thank you for your interest in contributing to the Alumet project!
+Thank you for your interest in contributing to Alumet!
 
-Here is what you need to know.
+Alumet is a collaborative project.
+Questions, suggestions, bug reports and pull requests are welcome.
 
-## The repositories
+To improve the experience of everyone (users, casual contributors, maintainers), please follow these guidelines.
 
-### Main repository (the one you are in)
+## How to ask a question
 
-This repository is divided in several parts:
-- The `alumet` crate contains the core of the measurement tool, as a Rust library.
-- Binaries can be created from this library, in order to provide a runnable measurement software. The official binaries that we provide are defined in `app-agent`. Agents always depend on `alumet`.
-- Plugins are defined in separate folders: `plugin-nvidia`, `plugin-rapl`, etc. Plugins always depend on `alumet`.
-- As an experimental feature, `alumet-ffi` contains a C API for building Alumet plugins.
-Folders `test-dynamic-plugins` and `test-dynamic-plugin-c` only exist to test this API
+Feel free to [open a discussion on GitHub](https://github.com/alumet-dev/alumet/discussions).
 
-### Other repositories
+## How to report a bug
 
-The [`alumet-dev` organization](https://github.com/alumet-dev) contains additional repositories for the website and the packaging of the tool. You'll find more information in each repository.
+1. Try to [find an issue](https://github.com/alumet-dev/alumet/issues?q=is%3Aissue) that already exists for your problem.
+2. If no one has reported your problem, open a [new issue](https://github.com/alumet-dev/alumet/issues). If you're unsure, open a [discussion](https://github.com/alumet-dev/alumet/discussions).
+We will convert it into an issue if needed.
 
-## What you can do
+## How to contribute to the project
 
-There are several categories of tasks that can help the Alumet project. You don't necessarily need to code in Rust!
+### Shared Values
 
-### Report issues
+Alumet is a free and open source software built by a variety of contributors.
+Together, we try to:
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,89 +1,235 @@
...
+```
 
-Manual action required: if you add a crate (like a plugin) to the repository, add the following two lines to your `Cargo.toml`:
+Example:
 
-```toml
-[lints]
-workspace = true
+```txt
+feat(plugins/csv): support colored output
```
titouanj, line 71: So the rule is to use the path to the plugin instead of its name?

titouanj (dismissed):

CONTRIBUTING.md
TheElectronWill: I believe that it's more clear to use a /-based hierarchy: `core/pipeline`, `plugins/csv`, etc. It does not have to match the file structure exactly. 
Do you think that it's better to use the crate name?

titouanj: No I don't have a specific opinion on this topic. I usually use the name of the component that is impacted by the commit, for example:
- ci(name of the workflow file) -> ci(main-pipeline)
- build(Makefile)
- fix(plugin-csv)

But if you believe that it's more clear to do like that, then I'm fully ok with it. But maybe you should add a paragraph to describe what the scope should be.

CONTRIBUTING.md
```diff
@@ -1,89 +1,249 @@
...
+
+List of common types: feat, fix, test, refactor, docs, ci, chore.
+
+The scope does not correspond to the Rust package, but adhere to the following format:
+
+```txt
+area/sub-area/optional-details
+```
+
+- When the "area" of change is small (e.g. the commit modifies the CI), the sub-area can be omitted. For instance, `fix(ci): disable flaky tests` is a valid commit message.
```
titouanj: what about the ci type? 🫣

titouanj: For me ci is the only thing that should not be a scope, but a type with the part of the ci that's fixed by the commit as a scope

CONTRIBUTING.md
TheElectronWill: I've clarified the scope format.

CONTRIBUTING.md
TheElectronWill: Ah, good point :D

CONTRIBUTING.md
```diff
@@ -1,89 +1,249 @@
...
+- For plugins, use `plugins/name`, without prefixing the plugin's name with `plugin-` (see the example above).
+- Additional levels (separated by /) are allowed.
+- List of common areas: core, plugins, agent, ci, readme
```
titouanj: Is readme for readme/csv? Or do I miss anything?
Because if it's what you have in mind, then why not using docs(plugin/csv) instead ?

CONTRIBUTING.md
```diff
@@ -1,89 +1,256 @@
...
+
+TL;DR:
+- Use the type system to your advantage. Make illegal states unrepresentable.
+- Write comments that explain the _why_, not the _how_. Add [documentation](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html) to modules, types and functions.
+- Use [anyhow](https://crates.io/crates/anyhow) (and call `.context` / `.with_context` for error messages) and [thiserror](https://crates.io/crates/thiserror).
+- When working on a plugin, let the Alumet framework guide you. If your code becomes super complicated, there might be a better way. Don't hesitate to (re)read the docs and to ask for help.
+- When working on the core, follow the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/about.html). Be careful of breaking changes. Strive to achieve great performance.
 
-When you add a new plugin to the repository, make it depend on the `alumet` crate in a relative way, without specifying a version. That is, the `dependencies` section of its `Cargo.toml` should look like this:
+Refer to the [developer book](https://alumet-dev.github.io/developer-book/) for more information.
```
titouanj: What about 
```suggestion
Refer to the [developer book][dev-book] for more information.
```

With the following at the bottom of the file:

```markdown
[dev-book]: https://alumet-dev.github.io/developer-book/ "ALUMET developer book"
```

TheElectronWill: good idea

CONTRIBUTING.md
```diff
@@ -1,89 +1,259 @@
...
+├── plugins/  -- official Alumet plugins
+└── separate-tests/ -- additional integration tests
+```
+
+#### Style
+
+TL;DR:
+- Use the type system to your advantage. Make illegal states unrepresentable.
+- Write comments that explain the _why_, not the _how_. Add [documentation](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html) to modules, types and functions.
+- Use [anyhow](https://crates.io/crates/anyhow) (and call `.context` / `.with_context` for error messages) and [thiserror](https://crates.io/crates/thiserror).
```
titouanj, line 171: Do we need to talk about: https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err ?

TheElectronWill: I think that this level of details should go in the developer book.
Furthermore, `thiserror` will take care of implementing `std::error::Error` properly.

titouanj (dismissed):

Issue #403, opened 2026-05-21
cpu_percent_% still higher than 100% in Alumet v0.9.3


thealanjason:
**Describe the bug**
We observe `cpu_percentage` still higher than 100% even though PR https://github.com/alumet-dev/alumet/pull/341 says that it was fixed in v0.9.3. 

The hardware on which we ran the tests has 96 CPU Cores. 

Refer the following two issues where we the bug surfaces: 
- https://github.com/thealanjason/alumet-viz/issues/7#issuecomment-4468320466
- https://github.com/thealanjason/topographic_uq_energy/issues/11
- https://github.com/thealanjason/topographic_uq_energy/pull/9#issuecomment-4320739915

Are we missing something? Do we need to use the `cgroups` plugin? 

**Environment information**
OS: Ubuntu Linux 24.04LTS
Alumet version²: v0.9.3.
Enabled plugins: 
```
    🧩 5 plugins started:
        - csv v0.2.0
        - rapl v0.3.1
        - procfs v0.2.1
        - nvml v0.5.0
        - energy-attribution v0.1.1
```

**To Reproduce**
Use the config here: https://github.com/thealanjason/energy_measurement/blob/main/measurement_tools/alumet/experiments/08_cpu_percentage_test/alumet-config-cpu-percent-test.toml

to `exec` the program here: https://github.com/thealanjason/energy_measurement/blob/86dcdd60bf89a0996294a7200fd82230c2658b3a/case_studies/gemm/gemm.py

The program would need a conda/micromamba to be created with the following `environment.yml` file
```
name: test-case-gemm
channels:
  - conda-forge
dependencies:
  - numpy
```

**Expected behavior**
`cpu_percent` less than 100% when the program uses all 96 cores.

**Logs / Output**
- Log file: https://github.com/thealanjason/energy_measurement/blob/86dcdd60bf89a0996294a7200fd82230c2658b3a/measurement_tools/alumet/experiments/08_cpu_percentage_test/alumet-agent-cpu-percent-test.log
- CSV Output file: https://github.com/thealanjason/energy_measurement/blob/86dcdd60bf89a0996294a7200fd82230c2658b3a/measurement_tools/alumet/experiments/08_cpu_percentage_test/alumet-output-rapl%2Benergy-attribution.csv

TheElectronWill:
[...]

PR #410, opened 2026-06-09
fix(plugin/procfs): cpu_percent higher than 100%

victoryeagle77:
Fix the issue #403

### DIFF
--- a/plugins/procfs/Cargo.toml
+++ b/plugins/procfs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plugin-procfs"
-version = "0.2.2"
+version = "0.3.0"
 edition.workspace = true
 repository.workspace = true
 
@@ -22,3 +22,4 @@ workspace = true
 alumet.workspace = true
 pretty_assertions.workspace = true
 toml.workspace = true
+tempfile.workspace = true
--- a/plugins/procfs/src/process.rs
+++ b/plugins/procfs/src/process.rs
@@ -64,6 +64,8 @@ struct ProcessStatsProbe {
 
     /// The memory page size, in bytes
     page_size: u64,
+
+    n_cores: usize,
 }
 
 impl ProcessStatsProbe {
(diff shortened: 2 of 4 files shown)


TheElectronWill (dismissed):

plugins/procfs/src/process.rs
```diff
@@ -414,7 +422,11 @@ impl ManualProcessMonitor {
         let ns_per_ticks = ns_per_ticks();
         Self {
             alumet_handle,
-            source_spawner: ProcessSourceSpawner { ns_per_ticks, metrics },
+            source_spawner: ProcessSourceSpawner {
+                ns_per_ticks,
+                metrics,
+                n_cores: num_cpus::get(),
```
ahuret: we already do the same thing in cgroups plugin, I think it make sense to align the method to get those.
(waiting for next step maybe using system_scan library)

https://github.com/alumet-dev/alumet/blob/main/plugins/cgroups/util-cgroups-plugins/src/cpus.rs#L40

ahuret (changes requested):

plugins/procfs/src/process.rs
victoryeagle77: Yes but it doesn't make sens to duplicate again this code part for the 3rd time (already written in cgroup AND rapl). For an other PR in the future we can eventually harmonize with a general/library file to distribute general functions like this between multiple plugins.

ahuret: I actually prefer to have 3 duplicates than 2 different methods of getting the same conceptual thing. This can lead to different behaviors and interpretation of measurements while using an unique method keep it consistent

ahuret: I would go with another duplicate and create an issue about factorizing it (which is probably something we could harmonize later with system_scan for example) - but at least for now everything will be aligned

victoryeagle77: Okay, if it's in the scope to harmonize later all `cpus.rs` calls I accept to do this.

ahuret (dismissed):

plugins/procfs/Cargo.toml
TheElectronWill: The version should be increased to `0.3.0`.

TheElectronWill (changes requested):

plugins/procfs/Cargo.toml
TheElectronWill: otherwise LGTM :)

victoryeagle77: Ah yes indeed

TheElectronWill (approved):

PR #414, opened 2026-06-23
Feat/brseed 1535 nvml clock info

Fyspeed:
Adds a new metric to NVML plugin, clock info. This metric measures clock frequency for each domain (Memory, Graphics, Video, SM).

Also updates "points_by_metric_and_consumer" in core/test to handle multiple attributes, + adds some tests for this function.


core/alumet/src/test/runtime.rs
```diff
@@ -711,10 +711,44 @@ impl<'a> SourceCheckOutputContext<'a> {
...
+    /// Output keys -> [
+    ///     ("clock_speed", "ResourceConsumer", [GPU]),
+    ///     ("clock_speed", "ResourceConsumer", [])
+    /// ]
+    /// ```
     pub fn points_by_metric_and_consumer(
         &'a self,
-        attribute_key: Option<&str>,
-    ) -> IndexMap<(&'a str, ResourceConsumer, Option<AttributeValue>), &'a MeasurementPoint> {
+        attribute_key_list: &[&str],
```
guigomcha: tbh the whole support for list thing seems a bit odd to me. Maybe you are trying to build a new function to get a point based on filters? I don't think that we can have multiple consumers for the same point.... Maybe I am not following a discussion that led to this or I am missing something

guigomcha: Just to give more context, I am pretty sure that having  attributes as lists in the metrics will give us a headache on several output plugins. For sure we'll need to look at the prometheus and opentelemetry exporters because labels have to be str:str key-value pairs... does this work in influxdb?

plugins/nvidia-nvml/src/probe.rs
```diff
@@ -306,6 +320,22 @@ impl<D: NvmlDevice> Source for FullSource<D> {
...
+                        timestamp,
+                        self.metrics.clock_info,
+                        self.resource.clone(),
+                        consumer.clone(),
+                        (device.clock_info(clock_type)? * 1000) as u64,
+                    )
+                    .with_attr("clock_type", clock_type_to_str(clock_type)?),
+                );
+            }
+        }
```
guigomcha, line 337: don't we get it in MHz?  I guess we'd have to multiply by 1e6?

guigomcha (changes requested):
awesome addition thanks!

core/alumet/src/test/runtime.rs
Fyspeed: This function is for testing only, and is currently only used for tests in NVML.

I do agree that this is a bit of a headache to set up .. basically, without this function, metrics with the same name but different attributes were not correctly handled (only one metric appeared in the IndexMap output instead of 1 per attribute value).
The whole approach should probably be reworked, but for now, i've just focused on making it work for my use case

Fyspeed: I guess I could remove entirely the "attribute_key"/"attribute_key_list" parameter ?

Also, another issue is that order is not guaranteed in the returned "Vec<AttributeValue>", as comparing AttributeValue seems complicated
So finding the right key may be tricky in some cases (for example, if the key is [attribute1, attribute2] in the IndexMap, but you try to find an item using key [attribute2, attribute1]...)

Idk if that part is clear

plugins/nvidia-nvml/src/probe.rs
Fyspeed: yep, fixing that thanks

core/alumet/src/test/runtime.rs
guigomcha: oook, my bad then. It just felt weird that the change was in the core but it is true that it is just a test

plugins/nvidia-nvml/src/nvml/features.rs
```diff
@@ -134,6 +144,14 @@ fn check_running_graphics_processes(device: &impl NvmlDevice) -> Result<Availabl
     )
 }
 
+/// Checks which version of `running_graphic_processes` is available (if any) on this NVML device.
```
ahuret: The comment seems not good

plugins/nvidia-nvml/README.md
```diff
@@ -44,6 +45,17 @@ The kind of the memory is the type of the allocated memory space reserved by the
 |`used`|Allocated device memory|
 |`reserved`|Device memory (in bytes) reserved for system use (driver or firmware)|
 
+#### Clock_type
+
+The speed of the clock may vary depending on the domain. These are the available domains:
```
ahuret: Shouldn't we rename this `clock_type` to `domain` then ? not sure

core/alumet/src/test/runtime.rs
```diff
@@ -711,10 +711,13 @@ impl<'a> SourceCheckOutputContext<'a> {
...
     }
 
+    /// ## Deprecated
+    /// Former utility function for tests for the NVML plugin.
+    /// Moved to plugins/nvidia-nvml/src/lib.rs
     pub fn points_by_metric_and_consumer(
         &'a self,
-        attribute_key: Option<&str>,
-    ) -> IndexMap<(&'a str, ResourceConsumer, Option<AttributeValue>), &'a MeasurementPoint> {
+        attribute_key_list: &[&str],
```
TheElectronWill: Changes here should be reverted, shouldn't they? Since we have decided to put it in the plugin itself.
Use the [deprecated attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)

TheElectronWill (changes requested):

my_config.toml
TheElectronWill, line 1: this file should not be part of your commits

TheElectronWill (changes requested):
almost done

TheElectronWill (approved):

PR #428, opened 2026-07-16
feat(plugin-perf): libpfm support

ahuret:


plugins/perf/README.md
```diff
@@ -41,6 +41,21 @@ To list the events that are available on your machine, run the `perf list` comma
 
 Note that based on your kernel version, some events could be unavailable.
```
AngeCyp, line 63: could you add a line or two about what is libpfm, it's goal,... ?

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,239 @@
...
+         set {LIBPFM_ENV} to the library's .so name or full path. Last error: {last_err}"
+    ))
+}
+
+/// Load libpfm once and initialize it. The result (success or failure) is cached for the
+/// whole program lifetime.
+fn libpfm() -> anyhow::Result<&'static LibPfm> {
+    static LIBPFM: OnceLock<Result<LibPfm, String>> = OnceLock::new();
+
+    let loaded = LIBPFM.get_or_init(|| {
```
AngeCyp: shouldn't we change the name since we can also recover from an error?
```suggestion
    let result = LIBPFM.get_or_init(|| {
```

AngeCyp (changes requested):

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        // SAFETY: the symbol signatures above match libpfm's C declarations.
+        unsafe {
+            let initialize: PfmInitializeFn = *lib
+                .get::<PfmInitializeFn>(b"pfm_initialize\0")
+                .map_err(|e| format!("symbol pfm_initialize not found: {e}"))?;
+            let strerror: PfmStrerrorFn = *lib
+                .get::<PfmStrerrorFn>(b"pfm_strerror\0")
+                .map_err(|e| format!("symbol pfm_strerror not found: {e}"))?;
+            let get_encoding: PfmGetEncodingFn = *lib
```
TheElectronWill: According to the man, this function is deprecated (since 2011!). Shouldn't we use `pfm_get_os_event_encoding`?

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,244 @@
...
+    unsafe { CStr::from_ptr(ptr).to_string_lossy().into_owned() }
+}
+
+/// Open the libpfm shared library, resilient to a non-standard name or location.
+///
+/// Resolution order:
+/// 1. `ALUMET_LIBPFM_LIB` — an explicit .so name (searched in the standard loader paths) or a
+///    full path to the shared object. Covers both a different name and a different location.
+/// 2. otherwise, the common .so names in [`LIBPFM_CANDIDATES`].
+fn open_library() -> Result<Library, String> {
```
TheElectronWill: Returning a proper anyhow Error would be more idiomatic, and it would allow to pass the error source properly (not inside the error string).

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,244 @@
...
+/// A handle to the dynamically-loaded libpfm, holding the function pointers we need.
+struct LibPfm {
+    // The library must stay alive as long as the function pointers are used. It lives in a
+    // `OnceLock` for the whole program, so it is never unloaded.
+    _lib: Library,
+    strerror: PfmStrerrorFn,
+    get_encoding: PfmGetEncodingFn,
+}
+
+// SAFETY: the raw function pointers point into `_lib`, which is kept alive alongside them.
```
TheElectronWill: But that does not make LibPfm thread-safe? I think that this is the wrong explanation here.

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,244 @@
...
+/// is derived from it with [`metric_suffix`].
+pub fn parse_event(name: &str) -> anyhow::Result<NamedPerfEvent<PfmEvent>> {
+    Ok(NamedPerfEvent {
+        name: metric_suffix(name),
+        description: format!("{name} (encoded via libpfm)"),
+        event: encode(name)?,
+    })
+}
+
+/// Resolve an event name (libpfm syntax, e.g. `RESOURCE_STALLS:ANY` or `INSTRUCTIONS:u`)
```
TheElectronWill: A link/pointer to `man libpfm` would be great. It describes the syntax.

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    #[test]
+    fn encode_generic_event() {
+        // The generic perf PMU events are available on every CPU that libpfm supports.
+        let event = encode("PERF_COUNT_HW_INSTRUCTIONS").expect("should encode a generic event");
+        // It should actually configure the attr when applied.
+        let mut attr = perf_event_attr::default();
+        event.update_attrs(&mut attr);
+        assert_eq!(attr.config, event.config);
```
TheElectronWill: Could we check more things here? To make sure that it's doing what we want.

```suggestion
        assert_eq!(attr.config, event.config);
        assert_ne!(attr.config, 0);
```

plugins/perf/src/pfm.rs
TheElectronWill: Anyhow has a nice `Context` for things like that: https://docs.rs/anyhow/latest/anyhow/trait.Context.html

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,244 @@
...
+    static LIBPFM: OnceLock<Result<LibPfm, String>> = OnceLock::new();
+
+    let result = LIBPFM.get_or_init(|| {
+        let lib = open_library()?;
+
+        // SAFETY: the symbol signatures above match libpfm's C declarations.
+        unsafe {
+            let initialize: PfmInitializeFn = *lib
+                .get::<PfmInitializeFn>(b"pfm_initialize\0")
+                .map_err(|e| format!("symbol pfm_initialize not found: {e}"))?;
```
TheElectronWill: You should use anyhow's Context in situations like this

plugins/perf/src/lib.rs
```diff
@@ -70,10 +71,17 @@ impl AlumetPlugin for PerfPlugin {
                 .map(|e| events::parse_cache(&e))
                 .try_collect()
                 .context("invalid cache event in config")?,
+            pfm_events: config
```
TheElectronWill: "pfm events" can also be hardware, software or cache events.

I think that it would be more clear for the user to have only one event list, and use the same approach as libpfm even when it's not present (with a much more limited support for events and modifiers, of course). For instance, we would have:
```
events = ["software::PAGE_FAULTS", "cache::BPU"]
```

plugins/perf/README.md
```diff
@@ -41,6 +42,17 @@ To list the events that are available on your machine, run the `perf list` comma
...
+**For libpfm events:**
+
+`perf_pfm_{event-name}` where `event-name` is any event supported by
+[libpfm4](https://perfmon2.sourceforge.net/) on your CPU, with non-alphanumeric characters
+replaced by `_` (e.g. `RESOURCE_STALLS:ANY` becomes the metric `perf_pfm_RESOURCE_STALLS_ANY`).
+
+**About libpfm:** The hardware/software/cache events above are only a small, generic set. A CPU actually has
+hundreds more events, but their codes are different on each CPU model. libpfm4 knows these
+codes: you give it a readable event name and it finds the right code for your CPU. To see the
+names available on your machine, run `showevtinfo` (from libpfm4) or `perf list`.
```
TheElectronWill: A paragraph about the libpfm event syntax?

TheElectronWill (changes requested):
I have some remarks to make the code more idiomatic, and one major remark [here](https://github.com/alumet-dev/alumet/pull/428#discussion_r3691358673) about user experience. We can talk about the consequences next week.

plugins/perf/src/pfm.rs
```diff
@@ -0,0 +1,256 @@
...
+    result.as_ref().map_err(|e| anyhow!("{e}"))
+}
+
+/// Resolve an event name (libpfm syntax, e.g. `RESOURCE_STALLS:ANY`) into a [`NamedPerfEvent`].
+/// Also validates that the event exists on the current CPU.
+/// A libpfm event name is structured like this (parsing is case-insensitive):
+///
+///`[pmu::]event_name[:unit_mask][:unit_mask…]`
+///
+///- **`pmu::`** *(optional)* : a libpfm PMU / microarchitecture model, e.g. `ix86arch::`, to
```
AngeCyp, line 160: `ix86arch` does it really exist ?

plugins/perf/src/spec.rs
```diff
@@ -127,130 +130,159 @@ struct Excludes {
...
 }
 
-/// Turn a string into a metric-name-safe suffix: non-alphanumeric characters become `_`, and
-/// leading/trailing `_` are trimmed.
-fn sanitize(s: &str) -> String {
+/// Turn a string into a metric-name-safe suffix: letters are lowercased, non-alphanumeric
+/// characters become `_`, and leading/trailing `_` are trimmed.
+pub(crate) fn sanitize(s: &str) -> String {
     let mapped: String = s
         .chars()
```
AngeCyp, line 270: If the String contains double space, it will get "_". Is it correct for your implementation or missing the reduce ?

AngeCyp (changes requested):

plugins/perf/src/pfm.rs
TheElectronWill: Actually, it does. See `man libpfm_intel_x86_arch` which reads "PMU name: ix86arch"

plugins/perf/src/spec.rs
TheElectronWill: I'm not sure I understand your comment. Don't we have this?
```rust
"__" => ""
"a  b" => "a__b"
"a b cµ" => "a_b_c"
```
which is ok for a metric name?

TheElectronWill (approved):

TheElectronWill (approved):

AngeCyp (approved):

PR #429, opened 2026-07-16
feat(plugin-perf): support configurable privileges

ahuret:
Current plugin perf always collect user+idle levels.

This change get it configurable, keeping the user+idle as default to not break compatibility.


TheElectronWill (commented):
The implementation seems OK, but I'm wondering whether this is the best approach.
If we implement a libpfm-like syntax for events, we could have a default level + a different privilege per event.

For a quicker solution, we could have a single value that specifies the default global level in a string, for instance `"k"` for kernel, `"uk"` for user+kernel, etc.

ahuret:
Closing because replaced by https://github.com/alumet-dev/alumet/pull/450

PR #430, opened 2026-07-17
feat(plugin-perf): compensate for counter multiplexing

ahuret:
Introduce a new attribute that give accuracy of the time series depending on whether or not the series had counters multiplexed.

Multiplexing happens when there is no more generic counter slot available.

The PR also introduce a new config that allow to compensate the counter value in case of multiplexing, by infering the value based on the time_running and time_enabled of the counter.

### DIFF
--- a/plugins/perf/README.md
+++ b/plugins/perf/README.md
@@ -43,6 +43,21 @@ Note that based on your kernel version, some events could be unavailable.
 
 ### Attributes
 
+Every measurement carries an `accuracy` attribute describing how faithful its value is (see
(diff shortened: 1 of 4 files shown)


TheElectronWill:
Good idea, it's important to offer this! I'm not sure if this is ready for review?

ahuret:
> Good idea, it's important to offer this! I'm not sure if this is ready for review?

it is now !

Aymeric-Brochier (commented):
LGTM

TheElectronWill (approved):

PR #434, opened 2026-07-18
fix(plugins/prometheus-exporter): register units on the right metrics

### LINKED ISSUE #433: prometheus-exporter: metrics with a unit lose it, unitless ones get a trailing underscore
**Describe the bug**
I think the unit-suffix branch in `plugins/prometheus-exporter/src/output.rs` is backwards. Right now metrics that have a unit go through plain `register()` and the ones without a unit go through `register_with_unit(Unit::Other("")`). It's flipped and you can see it on the wire 2 ways:

- metrics with a unit lose it: `kernel_cpu_time_alumet` comes out with no unit suffix and no `# UNIT` line (it should carry one, exemple `kernel_cpu_time_alumet_milliseconds` from the current `get_unit_string` logic)
- metrics without a unit grow a stray trailing underscore: `kernel_context_switches_alumet_`, `network_packets_alumet_`, and so on

Swapping the two arms fixes both, I've got a PR ready with a small regression test for it.

While I was in there, a smaller thing on the same plugin: the user book documents `append_unit_to_metric_name` and `use_unit_display_name` but the plugin's `Config` is `deny_unknown_fields` and has neither, so the documented example won't load as-is. I think those two keys actually belong to the csv plugin (they sit under `[plugins.csv]` in the conffile shipped with the `.deb`). A separate docs PR is coming for that.

**Environment information**
OS: Ubuntu 24.04 (arm) in a Docker container on an MacOS arm
Alumet version: 0.9.5 (`alumet-agent_0.9.5-1_arm64_ubuntu_24.04.deb`)
Enabled plugins: procfs and prometheus-exporter

**To Reproduce**
1. `alumet-agent --plugins procfs,prometheus-exporter run` (with a fresh config the packaged conffile trips #431 first)
2. `curl -s localhost:9091/metrics | grep -E '_alumet_?[{ ]'`

**Expected behavior**
Metrics with a unit keep it in the name plus a `# UNIT` line and metrics without one don't pick up a trailing underscore.

**Logs / Output**
```
kernel_cpu_time_alumet{...} 4950         # unit is ms, but the name doesn't show it
kernel_context_switches_alumet_{...} 42  # no unit, yet there's a trailing underscore
```
For context, I consume this exporter downstream and came to Alumet via a discussion thread, thanks for the pointer!

As for the issue #431 I posted earlier: _regarding the LLM policy, I used an AI assistant to help search existing issues, cross-check the plugin source and write up a part of this report (also for a better english). The bug and the reproduction come from my own test setup (a simulation lab in one of my projects that uses Alumet) and I understand the fix._

robintra:
Fixes #433

### DIFF
--- a/plugins/prometheus-exporter/src/output.rs
+++ b/plugins/prometheus-exporter/src/output.rs
@@ -69,8 +69,6 @@ impl alumet::pipeline::Output for PrometheusOutput {
         let mut registry = self.state.registry.blocking_write();
 
         for m in measurements {
-            let metric = ctx.metrics.by_id(&m.metric).unwrap();
-
             // Configure the name of the metric
             let full_metric = ctx
                 .metrics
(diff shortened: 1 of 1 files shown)


TheElectronWill:
Thanks for the PR!
Could you remove the merge commit and **rebase** on main instead? We use a rebase strategy to keep the history clean.

robintra:
Rebase done

TheElectronWill (approved):

PR #439, opened 2026-07-22
tests(core/watch): add unit tests for watch.rs file

AngeCyp:
Add tests for `watch.rs` file

### DIFF
--- a/core/alumet/src/agent/watch.rs
+++ b/core/alumet/src/agent/watch.rs
@@ -141,3 +141,102 @@ fn trigger_poll_now(pipeline: &MeasurementPipeline) -> anyhow::Result<()> {
         .block_on(send_task)
         .context("failed to send TriggerMessage")
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


core/alumet/src/agent/watch.rs
```diff
@@ -141,3 +141,90 @@ fn trigger_poll_now(pipeline: &MeasurementPipeline) -> anyhow::Result<()> {
...
+        child.wait().expect("Failed to wait child");
+    }
+
+    #[test]
+    fn test_wait_child_invalid_pid() {
+        let pid = 999999999;
+
+        let result = wait_child(pid);
+
+        assert!(result.is_ok());
```
titouanj: How can it be ok to wait for an invalid pid?

AngeCyp: because of implementation, if the process is too short for monitoring without having an explicit error when trying to watch, a non existant PID will not lead to an error

titouanj (approved):

PR #443, opened 2026-08-03
Implement basic version of temperature plugin

Mommessc:
This is a first version for a plugin to measure temperature of CPU packages and cores.
It relies on the [lm_sensors crate](https://docs.rs/lm-sensors/latest/lm_sensors/index.html) (Linux-only crate) and only looks at the `coretemp` feature, but other metrics could be retrieved such as fan speed later on.

Currently only works for Intel CPUs and tested on my personal laptop and a few Grid'5000 nodes.

I am planning to include a configuration option to retrieve the Package temperatures only (to avoid flooding the output with 16 CPU core measurements each time).

Disclaimer: this is the first Rust code I write, with a lot of mimic from other plugins.
I guess there is a lot of improvement for chained operations.
I also don't know how to properly deal with potential 'exceptions' throughout the code and how it integrates in Alumet.

Edit: there is absolutely no test right now but temperature values are not deterministic.

### DIFF
--- a/agent/src/bin/main.rs
+++ b/agent/src/bin/main.rs
@@ -59,6 +59,7 @@ fn load_plugins_metadata() -> Vec<PluginMetadata> {
             plugin_nvidia_jetson::JetsonPlugin,
             plugin_amd_gpu::AmdGpuPlugin<plugin_amd_gpu::AmdSmiProvider>,
             plugin_quarch::QuarchPlugin,
+            plugin_temperature::TemperaturePlugin,
         ]);
     }
 
--- a/plugins/temperature/src/lib.rs
+++ b/plugins/temperature/src/lib.rs
@@ -1,0 +1,152 @@
+use std::thread;
+use std::time::{Duration, Instant};
+
+use alumet::{
+        plugin::{
+            ConfigTable,
+            AlumetPluginStart,
+            rust::{AlumetPlugin, deserialize_config, serialize_config},
+        },
+        measurement::{MeasurementBuffer, MeasurementPoint, Timestamp},
+        units::Unit,
+        resources::{ResourceConsumer},
+};
+
+use serde::{Deserialize, Serialize};
+
+use lm_sensors::{Initializer, LMSensors};
+
+use crate::{
+    temperature_sensors::{SensorsFeature},
+};
+
+mod temperature_sensors;
+
+pub struct TemperaturePlugin {
+    config: Config,
+}
+
+#[derive(Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct Config {
+    // TODO: consider adding an option to use Fahrenheit degrees instead of Celsius
+    
+    /// Initial interval between two temperature measurements.
+    #[serde(with = "humantime_serde")]
+    pub poll_interval: Duration,
+
+    /// Initial interval between two flushing of measurements.
+    #[serde(with = "humantime_serde")]
+    pub flush_interval: Duration,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            poll_interval: Duration::from_secs(1), // 1Hz
+            flush_interval: Duration::from_secs(5),
+        }
+    }
+}
+
+
+impl AlumetPlugin for TemperaturePlugin {
+    fn name() -> &'static str {
+        "temperature"
+    }
+
+    fn version() -> &'static str {
+        env!("CARGO_PKG_VERSION")
+    }
+
+    fn default_config() -> anyhow::Result<Option<ConfigTable>> {
+        let config = serialize_config(Config::default())?;
+        Ok(Some(config))
+    }
+
+    fn init(config: ConfigTable) -> anyhow::Result<Box<Self>> {
+        let config = deserialize_config(config)?;
+        Ok(Box::new(TemperaturePlugin { config }))
+    }
+
+    fn start(&mut self, alumet: &mut AlumetPluginStart) -> anyhow::Result<()> {
+        // Create the metric
+        let lm_metric = alumet.create_metric::<f64>(
+            "temperature_lm_sensor",
+            Unit::DegreeCelsius,
+            "Temperature measured, as reported by lm_sensors."
+        )?;
+
+        // Get the config parameters out of 'self', to be used in the thread
+        let poll_interval = self.config.poll_interval;
+
+        // A flush of the measurement buffer will be done after `flush_rounds` measurements
+        let flush_rounds = ((self.config.flush_interval.as_nanos() / poll_interval.as_nanos()) as usize).max(1);
+        
+        // Create an autonomous source and add the source to the measurement pipeline
+        alumet.add_autonomous_source_builder("temperature_autonomous_source", move |_ctx, cancel_token, tx| {
+            let out_tx = tx.clone();
+
+            // An new thread is needed because LMSensors and internal objects
+            // are not Sync nor Send
+            let handle = thread::spawn(move || {
+                // Initialise all the things related to lm_sensors
+                // TODO: fail gracefully instead of calling 'unwrap'
+                let lmsensors: LMSensors = Initializer::default().initialize().unwrap();
+                
+                // Only getting temperature sensors from coretemp.
+                let sensors_feature_list: Vec<SensorsFeature> = temperature_sensors::get_coretemp_feature_list(&lmsensors);
+                log::info!("Got {} temperature sensor features", sensors_feature_list.len());
+                
+                let mut buf = MeasurementBuffer::new();
+                let mut next_poll_time = Instant::now() + poll_interval;
+                let mut round = 0;
+
+                while !cancel_token.is_cancelled() {
+                    // Get measurement from all temperature sensors
+                    for feature in &sensors_feature_list {
+                        //TODO: fail gracefully instead of calling 'unwrap'
+                        let temperature = feature.read_temperature_value().unwrap();
+
+                        buf.push(MeasurementPoint::new(
+                            Timestamp::now(),
+                            lm_metric,
+                            feature.get_resource(),
+                            ResourceConsumer::LocalMachine,
+                            temperature
+                        ));
+                    }
+                    
+                    round += 1;
+                    if round == flush_rounds {
+                        // Push and clear the buffer after at most the flush_interval period
+                        let _ = out_tx.blocking_send(buf.clone());
+                        buf.clear();
+                        round = 0;
+                    }
+
+                    thread::sleep(next_poll_time - Instant::now());
+                    next_poll_time += poll_interval;
+                }
+
+                // Flush one last time the buffer
+                let _ = out_tx.blocking_send(buf.clone());
+                buf.clear();
+            });
+
+            let source = Box::pin(async move {
+                // Just wait for the thread to terminate
+                handle.join().unwrap();
+
+                Ok(())
+            });
+            Ok(source)
+        }).expect("source names should be unique (in the same plugin)");
+
+        Ok(())
+    }
+
+    fn stop(&mut self) -> anyhow::Result<()> {
+        Ok(())
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,6 +110,7 @@ dependencies = [
  "plugin-relay",
  "plugin-slurm",
  "plugin-socket-control",
+ "plugin-temperature",
  "pretty_assertions",
  "serde",
  "tempfile",
@@ -420,6 +421,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "bindgen"
+version = "0.72.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
+dependencies = [
+ "bitflags 2.13.0",
+ "cexpr",
+ "clang-sys",
+ "itertools 0.13.0",
+ "log",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex 1.3.0",
+ "syn",
+]
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -574,7 +595,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e228eec9be7c17ccb640b59b36a5cd805ea2a564a4c5e162c2f659fea30d3b96"
 dependencies = [
  "find-msvc-tools",
- "shlex",
+ "shlex 2.0.1",
+]
+
+[[package]]
+name = "cexpr"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
+dependencies = [
+ "nom",
 ]
 
 [[package]]
@@ -638,6 +668,17 @@ checksum = "57663b653d948a338bfb3eeba9bb2fd5fcfaecb9e199e87e1eda4d9e8b240fd9"
 dependencies = [
  "ciborium-io",
  "half",
+]
+
+[[package]]
+name = "clang-sys"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "157a8ba7b480713b56f4c09fd13fc3e0a22a5dfab8097ba61cbc5feef950788a"
+dependencies = [
+ "glob",
+ "libc",
+ "libloading",
 ]
 
 [[package]]
@@ -1159,6 +1200,12 @@ name = "dtoa"
 version = "1.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4c3cf4824e2d5f025c7b531afcb2325364084a16806f6d47fbc1f5fbd9960590"
+
+[[package]]
+name = "dunce"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "either"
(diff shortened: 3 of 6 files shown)


plugins/temperature/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+                    
+                    round += 1;
+                    if round == flush_rounds {
+                        // Push and clear the buffer after at most the flush_interval period
+                        let _ = out_tx.blocking_send(buf.clone());
+                        buf.clear();
+                        round = 0;
+                    }
+
+                    thread::sleep(next_poll_time - Instant::now());
```
AngeCyp: If we are late, we will try to sleep for a negative amount of time
I suggest something like:
```suggestion
                    thread::sleep(next_poll_time.saturating_duration_since(Instant::now()));
```

plugins/temperature/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+                        //TODO: fail gracefully instead of calling 'unwrap'
+                        let temperature = feature.read_temperature_value().unwrap();
+
+                        buf.push(MeasurementPoint::new(
+                            Timestamp::now(),
+                            lm_metric,
+                            feature.get_resource(),
+                            ResourceConsumer::LocalMachine,
+                            temperature
+                        ));
```
AngeCyp: As you wrote, unwrap is kind of ugly, maybe you could use a match case to log the error if you don't want to exit for an error, or to handle correctly the error
```suggestion
                        let temperature = feature.read_temperature_value().unwrap();
                        match feature.read_temperature_value() {
                                            Ok(temperature) =>  {buf.push.................. },
                                            Err(e) => log::warn!("failed to read {:?}: {e}", feature),
                            }
```

plugins/temperature/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                    
+                    round += 1;
+                    if round == flush_rounds {
+                        // Push and clear the buffer after at most the flush_interval period
+                        let _ = out_tx.blocking_send(buf.clone());
+                        buf.clear();
```
AngeCyp: Clone then clear the buffer, maybe there is a optimization to find here

plugins/temperature/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        // Get the config parameters out of 'self', to be used in the thread
+        let poll_interval = self.config.poll_interval;
+
+        // A flush of the measurement buffer will be done after `flush_rounds` measurements
+        let flush_rounds = ((self.config.flush_interval.as_nanos() / poll_interval.as_nanos()) as usize).max(1);
+        
+        // Create an autonomous source and add the source to the measurement pipeline
+        alumet.add_autonomous_source_builder("temperature_autonomous_source", move |_ctx, cancel_token, tx| {
+            let out_tx = tx.clone();
```
AngeCyp: tx is not used anywhere in the closure, you can avoid cloning it and directly use `tx` instead of `out_tx`

plugins/temperature/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+                }
+
+                // Flush one last time the buffer
+                let _ = out_tx.blocking_send(buf.clone());
+                buf.clear();
+            });
+
+            let source = Box::pin(async move {
+                // Just wait for the thread to terminate
+                handle.join().unwrap();
```
AngeCyp: Could this return an error ? 
If yes, maybe prefer a match case or something like this

AngeCyp (changes requested):
Thanks for your submission !

AngeCyp:
Don't hesitate to run: 
```bash
cargo fmt
```
on your code to apply rust formatting.

You can also use clippy
```
cargo-clippy
``` 

To see if you code contains errors. (Don't worry, many errors will output on other part of the code for the moment)

plugins/temperature/src/lib.rs
Mommessc: I simply followed the pseudo-doc for the autonomous source creation in the code of the builder (https://github.com/alumet-dev/alumet/blob/main/core/alumet/src/plugin/phases.rs#L231), but I agree there is probably a better solution.

plugins/temperature/src/lib.rs
Mommessc: What about `handle.join()?;` and let Alumet deal with errors from the source?

Mommessc:
Just pushed new commits to include the main plugin file (I forgot to add it in first batch of commits ...) and with the modifications you suggested.
I also added a plugin configuration to filter out some sensors.

plugins/lm-sensors/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+            // An new thread is needed because LMSensors and internal objects
+            // are not Sync nor Send
+            let handle = thread::spawn(move || {
+                // Initialise all the things related to lm_sensors
+                let lmsensors: LMSensors = Initializer::default()
+                    .initialize()
+                    .expect("Could not initialise LMSensors.");
+
+                // This plugin only works for Intel processors for the moment
+                // TODO: check the CPU vendor and stop if it is not Intel
```
Mommessc, line 114: There is already code in the RAPL plugin to do this by parsing output of `lscpu`, but I am not sure how to proceed here.

AngeCyp (changes requested):
Can you add a README for your plugin. You can look for example there: [grace-hopper README](https://github.com/alumet-dev/alumet/blob/main/plugins/grace-hopper/README.md)
Don't hesitate to look for other README

TheElectronWill:
We have a template for the plugins' readme: See [SOURCE_PLUGIN_README_TEMPLATE.md](https://github.com/alumet-dev/alumet/blob/main/readme/SOURCE_PLUGIN_README_TEMPLATE.md)

agent/src/bin/main.rs
```diff
@@ -59,6 +59,7 @@ fn load_plugins_metadata() -> Vec<PluginMetadata> {
             plugin_nvidia_jetson::JetsonPlugin,
             plugin_amd_gpu::AmdGpuPlugin<plugin_amd_gpu::AmdSmiProvider>,
             plugin_quarch::QuarchPlugin,
+            plugin_temperature::TemperaturePlugin,
```
ahuret: I would rather call the plugin `LmSensorsPlugin`, which is the tech providing temperature and other metrics. The semantic of plugins are based on the tech that bring metrics, because multiple plugins can bring the same metrics for different stacks (eg: we could have another plugin using LibreHardwareMonitor to collect temperature for Windows).

(this apply to other code snippets that talk about "temperature" instead of lm_sensors)

TheElectronWill: Good point, I was about to publish the same comment :smile:

Mommessc:
Just renamed to `LMSensorsPlugin`.

For now I did not modify the internal code of the plugin that assumes only the temperature measurements are requested.
In the plugin configuration I added a field to enable the temperature measurements, in a tentative to be future-proof-retro-compatible if more metrics are added to this plugin.

plugins/lm-sensors/src/temperature.rs
```diff
@@ -0,0 +1,96 @@
...
+        "Package" => {
+            if let Ok(id) = v.get(2).expect("Expected Package id value").parse() {
+                return Ok(Resource::CpuPackage { id });
+            }
+        }
+        "Core" => {
+            if let Ok(id) = v.get(1).expect("Expected Core id value").parse::<u32>() {
+                // Need to attach the package id to the CPU core id
+                let custom_id = format!("{}_{}", coretemp_id, id);
+                return Ok(Resource::Custom {
```
ahuret, line 46: no need for custom, there's already a CpuCore resource (see [here](https://github.com/alumet-dev/alumet/blob/main/core/alumet/src/resources.rs#L52))

plugins/lm-sensors/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            // General plugin default configuration
+            poll_interval: Duration::from_secs(1), // 1Hz
+            flush_interval: Duration::from_secs(5),
+
+            // Temperature default configuration
+            enable_temperature: false,
```
ahuret, line 57: I would have enable it by default, even more as this is the only feature in the plugin for now

plugins/lm-sensors/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+    fn start(&mut self, alumet: &mut AlumetPluginStart) -> anyhow::Result<()> {
+        if !self.config.enable_temperature {
+            return Err(anyhow!(
+                "For the moment only the temperature feature of LMSensors is implemented.\nPlease enable it in the configuration with the 'enable_temperature' field."
+            ));
+        }
+
+        // Create the metric
+        let temperature_metric = alumet.create_metric::<f64>(
+            "lm_sensors_temperature",
```
ahuret, line 91: I'm advocating for calling it just "temperature". The fact that it's coming from lm_sensors shouldn't be reported in the name imo (I know it's something we did a lot in other plugins, but we're changing it).
And as we have the resource attached to the metric, we know the temperature of what it is (the temperature of cpu_core for ex). Note that we also have an initiative in // to annotate the source of metrics automatically.

plugins/lm-sensors/src/temperature.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    // If we reach this line we could not parse the feature name
+    Err(anyhow!("Failed to parse the sensor feature name"))
+}
+
+pub fn get_coretemp_sensors_list<'a>(lmsensors: &'a LMSensors, package_only: bool) -> Vec<SensorsFeature<'a>> {
+    let mut temperature_sensors_list: Vec<SensorsFeature> = vec![];
+    for chip in lmsensors.chip_iter(None).filter(|x| {
+        x.name()
+            .expect("Chip name from LMSensors is not a valid UTF-8 string.")
```
ahuret, line 63: return errors instead of expect (let the caller handle the error)

ahuret: same for other expect in the function

plugins/lm-sensors/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        // Get the config parameters out of 'self', to be used in the thread
+        let poll_interval = self.config.poll_interval;
+        let coretemp_package_only = self.config.coretemp_package_only;
+
+        // A flush of the measurement buffer will be done after `flush_rounds` measurements
+        let flush_rounds = ((self.config.flush_interval.as_nanos() / poll_interval.as_nanos()) as usize).max(1);
+
+        // Create an autonomous source and add the source to the measurement pipeline
+        alumet.add_autonomous_source_builder("lm_sensors_autonomous_source", move |_ctx, cancel_token, tx| {
```
ahuret, line 104: isn't it possible to make a "normal" source ?

plugins/lm-sensors/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+alumet.workspace = true
+anyhow.workspace = true
+log.workspace = true
+humantime-serde.workspace = true
+serde = { workspace = true, features = ["derive"] }
+sensors-sys = "0.2.18"
```
ahuret, line 13: this library force us to have libsensors-dev installed at build time. Which means a new dependency for our CI pipeline at least. Few options:

- Keep it like this (maintain the dependency to libsensors-dev and document it)
- Make the plugin lm-sensors optional using cargo features, meaning it's the user choice to bring this plugin or not
- Get rid of sensors-sys crate, load the library dynamically at runtime (require to implement) -> this is what we do in nvml plugin or perf plugin btw

wdyt @TheElectronWill  ?

plugins/lm-sensors/src/temperature.rs
ahuret: As I understand, the `custom_id` here has the format `{cpu_package_id}_{cpu_core_id}`.
If you keep only the cpu_core_id part, it will perfectly fit the CpuCore{ id } schema

plugins/lm-sensors/src/temperature.rs
```diff
@@ -0,0 +1,96 @@
+use anyhow::anyhow;
+
+use lm_sensors::{FeatureRef, LMSensors, SubFeatureRef};
+
+use alumet::resources::Resource;
+
+pub struct SensorsFeature<'a> {
```
ahuret, line 7:
```suggestion
pub struct CoretempFeature<'a> {
```

plugins/lm-sensors/src/temperature.rs
ahuret, line 1: I propose to rename it `coretemp.rs`.

plugins/lm-sensors/src/temperature.rs
ahuret: or CoretempSensor

ahuret (changes requested):

plugins/lm-sensors/src/temperature.rs
Mommessc: A custom id is required to attach the package id information as well.
On a machine with two CPU packages the different cores will be identified as `0_0`, `0_1`, `0_2`... for the first CPU and `1_0`, `1_1`, `1_2` ... for the second CPU.

With the `CpuCore{ id }` schema there would be two occurrences of each CPU core id (one for each CPU package) without the possibility to distinguish.
Maybe there is another way to do so in Alumet I am not aware of.

plugins/lm-sensors/src/lib.rs
Mommessc: If you refer to `lm_sensors_temperature` I don't think it is a good idea to call it just `temperature`.
Metric names are supposed to be unique ([from developer's book](https://github.com/alumet-dev/developer-book/blob/main/src/plugins/metrics_measurements.md)) and to me it makes sense to write from where the temperautre comes from.

I can imagine a scenario where one wants to get temperature measurements of CPUs from different plugins, so if all plugins able to measure some kind of temperature report their metric as just `temperature` there will be collisions and no mean to distinguish between measurements from two different plugins.

plugins/lm-sensors/src/lib.rs
Mommessc: Unfortunately no :(
`LMSensors` (and other things in the crate) [are not Send](https://docs.rs/lm-sensors/latest/lm_sensors/struct.LMSensors.html#impl-Send-for-LMSensors), which is required for a normal source.

After a lot of head-scratching and discussion with @TheElectronWill we saw two options: either use an autonomous source with a dedicated thread to handle `LMSensors`, or don't use the crate at all and kind of re-invent the wheel by reading the `hwmon` temperature files by hand.
We chose the first option (and that would make an example of how to use autonomous sources in Alumet, I haven't found examples for that).

plugins/lm-sensors/Cargo.toml
Mommessc: Making the plugin optional seems a good solution to me, but I have absolutely no idea how do this.
Help requested on this.

plugins/lm-sensors/src/coretemp.rs
```diff
@@ -53,44 +53,53 @@ fn resource_from_string(name: &str, coretemp_id: u32) -> anyhow::Result<Resource
...
+                                     |v| v.starts_with("Package"))
+            })
+            .map(|x| CoretempSensor::new(x, coretemp_id))
+            // If CoretempSensor constructor fails, the Err is propagated in the collect
+            .collect();
+
+        match tmp_sensors_list {
+            Ok(tmp_sensors_list) => coretemp_sensors_list.extend(tmp_sensors_list),
+            Err(e) => return Err(anyhow!("Could not create coretemp sensor: {e}"))
+        }
```
Mommessc: This is the least ugly solution I found to remove expect calls and correctly return errors.
If there is any other Rust-friendly way of doing it, I would be happy to know (especially for the map_or_else thingy where I wanted to return the error directly inside the arm dealing with an error).

Mommessc:
Just rebased from `main` to be up-to-date so only the last 6 commits are new from last time.

plugins/lm-sensors/Cargo.toml
TheElectronWill: I agree. It should dynamically load the library at runtime. Do you need help doing so @Mommessc?

Mommessc: I will try to mimic what is done in the nvml or perf plugins to dynamically load a library.
If this is not sufficient I will get back to ask for help

Mommessc: Okay so it turns out the plugin does not need the `sensors-sys` crate dependency because it's already a dependency of the `lm-sensors` crate.

Playing with a docker container (from the `rust:latest` image) I was able to build alumet in release mode by manually installing the `clang` and `libsensors-dev` packages (see update in the README for more details).

Then with another docker container with the same image I was able to execute the `alumet-agent` binary by manually installing the `libsensors5` package, using in the config only the CSV and lm-sensors plugins.
BUT it turns out trying to execute alumet-agent with other plugins also requires the `libsensors5` package.
So I guess that means even if the lm-sensors plugin is not used, the alumet-agent needs at runtime the `libsensors5` library.

From my understanding, the `lmsensors-sys` crate should dynamically load `libsensors` at runtime.
And I thought all the alumet plugins were somewhat loaded/activated at runtime by the alumet-agent only if they were used, but apparently this is not true.
When trying to execute alumet-agent without the lm-sensors plugin in the config and without `libsensors5` installed in the environment the execution crashed.
This is where my knowledge stops and I hope you can help me out in this.

plugins/lm-sensors/src/lib.rs
TheElectronWill: Not yet. Because "normal" task scheduling is based on tokio, we require a `Send` bound on every source, and it doesn't work with libsensors.

plugins/lm-sensors/src/lib.rs
ahuret: Actually, it will be possible to make it when we will have implemented the Origin/Source concept onto MeasurementPoint. Which is not the case now, but anyway we still prefer keep an abstracted semantic.
If you want you can add an attribute (`origin=lm-sensors`) to your point waiting for that feature ?

ahuret: Imagine this scenario:

You have a Grafana dashboard showing the temperature of you CPUs, everytime you need to change the technology that measure the temperature under the hood, you need to change the dashboard definition to match the new metric names.
Same problem if you make model on top of metrics, the formula of the model shouldn't care about the underlying technology, it should care about a clear semantic and unit and should be ready to work with new plugins/techno without recoding the thing

plugins/lm-sensors/src/coretemp.rs
```diff
@@ -0,0 +1,121 @@
...
+                .parse::<u32>()
+            {
+                // Need to attach the package id to the CPU core id
+                let custom_id = format!("{}_{}", coretemp_id, id);
+                return Ok(Resource::Custom {
+                    kind: std::borrow::Cow::Borrowed("cpu_core"),
+                    id: custom_id.into(),
+                });
+            }
+        }
```
ahuret, line 59: About that, we had the discussion about using the CpuCore instead. Discussed with Guillaume, the issue is that we had the wrong definition of the cpu_core_id. We added more definition of what is a CpuCore.id here: https://github.com/alumet-dev/alumet/blob/main/core/alumet/src/resources.rs#L53

So instead of having custom resource, you should use CpuCore applying this new cpu_core.id meaning

PR #444, opened 2026-08-04
feat(cgroups): add io pressure in cgroups

AngeCyp:
Add a new metric about io pressure. (how much time processes are delayed because they are waiting for disk or other IO resources to become available)
Read the file: `io.pressure` on cgroup.

2 new metrics:

- `io_pressure_full_total`
- `io_pressure_some_total`

Also add tests

### DIFF
--- a/plugins/cgroups/raw/README.md
+++ b/plugins/cgroups/raw/README.md
@@ -19,6 +19,8 @@ Here are the metrics collected by the plugin's sources.
 |`cgroup_memory_file`|Gauge|Bytes|memory used to cache filesystem data|`LocalMachine`|`Cgroup`|see below|
 |`cgroup_memory_kernel_stack`|Gauge|Bytes|memory allocated to kernel stacks|`LocalMachine`|`Cgroup`|see below|
 |`cgroup_memory_pagetables`|Gauge|Bytes|memory reserved for the page tables|`LocalMachine`|`Cgroup`|see below|
+|`io_pressure_some_total`|CounterDiff|microseconds|IO pressure some total delta (at least one task stalled)|`LocalMachine`|`Cgroup`|none|
+|`io_pressure_full_total`|CounterDiff|microseconds|IO pressure full total delta (all tasks stalled)|`LocalMachine`|`Cgroup`|none|
 
 ### Attributes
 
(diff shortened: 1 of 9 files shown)


plugins/cgroups/raw/README.md
```diff
@@ -19,6 +19,8 @@ Here are the metrics collected by the plugin's sources.
 |`cgroup_memory_file`|Gauge|Bytes|memory used to cache filesystem data|`LocalMachine`|`Cgroup`|see below|
 |`cgroup_memory_kernel_stack`|Gauge|Bytes|memory allocated to kernel stacks|`LocalMachine`|`Cgroup`|see below|
 |`cgroup_memory_pagetables`|Gauge|Bytes|memory reserved for the page tables|`LocalMachine`|`Cgroup`|see below|
+|`io_pressure_some_total`|CounterDiff|Unity|IO pressure some total delta (at least one task stalled)|`LocalMachine`|`Cgroup`|none|
+|`io_pressure_full_total`|CounterDiff|Unity|IO pressure full total delta (all tasks stalled)|`LocalMachine`|`Cgroup`|none|
```
Aymeric-Brochier: I think Unit should be microseconds :
and have this in the Description: The total absolute stall time

https://docs.kernel.org/accounting/psi.html

"The ratios (in %) are tracked as recent trends over ten, sixty, and three hundred second windows, which gives insight into short term events as well as medium and long term trends. The total absolute stall time (in us) is tracked and exported as well, to allow detection of latency spikes which wouldn’t necessarily make a dent in the time averages, or to average trends over custom time frames."

Aymeric-Brochier (commented):
Only checked README.md

plugins/cgroups/util-cgroups-plugins/src/metrics.rs
```diff
@@ -104,6 +112,16 @@ impl Metrics {
...
+        let io_pressure_some_total = alumet.create_metric::<u64>(
+            "io_pressure_some_total",
+            PrefixedUnit::micro(Unit::Second),
+            "IO pressure some total delta: time with at least one task stalled since previous measurement",
+        )?;
+        let io_pressure_full_total = alumet.create_metric::<u64>(
+            "io_pressure_full_total",
+            PrefixedUnit::micro(Unit::Second),
+            "IO pressure full total delta: time with all tasks stalled since previous measurement",
+        )?;
```
ahuret, line 124: shouldn't we create those metrics only if they are activated (in settings) ?

AngeCyp: The metrics are always created for now. A config exist and so later we should works to make a filter about which metric we really want. For now it's hardcoded

ahuret (approved):

titouanj (approved):

PR #451, opened 2026-08-10
feat(cgroup): add max memory metric

AngeCyp:
- add max memory metric reading `memory.max` file.
  - if *max* is contained in the file, write the maximum of ram as limit 
- add tests around this new metric


plugins/cgroups/util-cgroups/src/measure/parse.rs
```diff
@@ -34,6 +34,90 @@ pub unsafe fn parse_single_u64(io_buf: &[u8]) -> io::Result<u64> {
...
+
+    Err(io::Error::new(io::ErrorKind::InvalidData, "MemTotal not found"))
+}
+
+/// Reads and parses the total system memory from `/proc/meminfo`.
+///
+/// This function reads the entire `/proc/meminfo` file and extracts the MemTotal value,
+/// which represents the total amount of physical RAM available in the system.
+///
+/// # Returns
```
TheElectronWill, line 78: The usual way of documenting errors is with an `# Errors` section. See the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure).

plugins/cgroups/util-cgroups/src/measure/parse.rs
```diff
@@ -34,6 +34,90 @@ pub unsafe fn parse_single_u64(io_buf: &[u8]) -> io::Result<u64> {
...
+/// * `Ok(u64)` - The total system memory in bytes
+/// * `Err(io::Error)` - An I/O error if:
+///   - The file cannot be read
+///   - The file content cannot be parsed
+///   - MemTotal is not found in the file
+///
+/// # Platform
+///
+/// This function is Linux-specific as it relies on the `/proc/meminfo` filesystem,
+/// which is a Linux kernel interface for system information.
```
TheElectronWill: Should it be guarded by a [`#[cfg(target_os = …)]](https://doc.rust-lang.org/reference/conditional-compilation.html#target_os) then? Or not because cgroupfs only exist on Linux, in which case you don't need to include that info here?

plugins/cgroups/util-cgroups/src/measure/parse.rs
```diff
@@ -34,6 +34,90 @@ pub unsafe fn parse_single_u64(io_buf: &[u8]) -> io::Result<u64> {
...
+
+/// Parses a single `u64` value or the string "max" from `io_buf`.
+///
+/// Returns total available if the content is "max", otherwise parses as u64.
+///
+/// # Safety
+/// The bytes passed in must be valid UTF-8.
+pub unsafe fn parse_single_u64_or_max(io_buf: &[u8], maximum_value: &mut Option<u64>) -> io::Result<u64> {
+    let content = unsafe { std::str::from_utf8_unchecked(io_buf.trim_ascii_end()) };
+    let trimmed = content.trim();
```
TheElectronWill: Why trim twice?

TheElectronWill (changes requested):

plugins/cgroups/util-cgroups/src/measure/parse.rs
AngeCyp: No reason :/

titouanj (approved):
LGTM

TheElectronWill (approved):

PR #455, opened 2026-08-13
Feat/add swap metrics in cgroups

AngeCyp:
Add two new metrics about swap current usage on cgroup and swap maximum available.

- `cgroup_memory_swap_current`
- `cgroup_memory_swap_max`

Also add unit tests. And modify the `ValueFile` structure to better handle of read in case of **max**.


plugins/cgroups/util-cgroups/src/measure/parse.rs
```diff
@@ -89,29 +125,52 @@ fn parse_memtotal(meminfo: &str) -> io::Result<u64> {
...
+/// - The value represents the total swap space configured on the system
+/// - If no swap is configured, this function will return Ok(0)
+pub fn total_swap() -> io::Result<u64> {
+    let meminfo = fs::read_to_string("/proc/meminfo")?;
+    parse_swaptotal(&meminfo)
+}
+
 /// Parses a single `u64` value or the string "max" from `io_buf`.
 ///
 /// Returns total available if the content is "max", otherwise parses as u64.
```
ahuret: change the description to mention the new behavior

plugins/cgroups/util-cgroups/src/measure/parse.rs
```diff
@@ -89,29 +125,52 @@ fn parse_memtotal(meminfo: &str) -> io::Result<u64> {
...
 ///
 /// # Safety
 /// The bytes passed in must be valid UTF-8.
-pub unsafe fn parse_single_u64_or_max(io_buf: &[u8], maximum_value: &mut Option<u64>) -> io::Result<u64> {
+pub unsafe fn parse_single_u64_or_max(io_buf: &[u8]) -> io::Result<Option<u64>> {
```
ahuret: Using an enum like 
```rust
enum U64MaxResult {
  U64(u64),
  Max,
}
```
instead of having the Ok(None) being the "max"
would be better imo (could be called ValueFileResult as well)

ahuret (changes requested):

ahuret (approved):

PR #456, opened 2026-08-14
Ci/add cargo machete

Aymeric-Brochier:


.github/workflows/rust_ci.yml
```diff
@@ -23,10 +23,25 @@ jobs:
           components: rustfmt
       - name: Check formatting
         run: cargo fmt --check
+  check_dependencies:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v7
+        - uses: dtolnay/rust-toolchain@master
```
AngeCyp, line 30: Not fan of master. In it's example stable is used. Is there another tag usable ?

AngeCyp (changes requested):

.github/workflows/rust_ci.yml
Aymeric-Brochier: The tag "master" is also used in other jobs like check_format 
We can leave it for now and plan to change all of them at once in another PR

AngeCyp (approved):

PR #457, opened 2026-08-14
Perf raw hex

ahuret:
Adding a new way to configure event using hexadecimal notation.

It's based on top of https://github.com/alumet-dev/alumet/pull/428 so need to be merged after


TheElectronWill (approved):
lgtm

PR #458, opened 2026-08-19
feat(raw-cgroup): modify the way to handle max capping

AngeCyp:
modify the way to handle max capping on memory and swap files.

Now, both metrics are `F64` instead of `U64` and return **-1**. It's because, it's not meaningful to return all available Swap or RAM as it is shared. 

Updated README according to this new add


plugins/cgroups/raw/README.md
```diff
@@ -12,9 +12,10 @@ Here are the metrics collected by the plugin's sources.
 
 |Name|Type|Unit|Description|Resource|ResourceConsumer|Attributes|
 |----|----|----|-----------|--------|----------------|----------|
-|`cpu_time_delta`|CounterDiff|nanoseconds|time spent by the pod executing on the CPU|`LocalMachine`|`Cgroup`|see below|
+|`cpu_time_delta`|CounterDiff|nanoseconds|time spent by the cgroups executing on the CPU|`LocalMachine`|`Cgroup`|see below|
 |`cpu_percent`|Gauge|Percent (0 to 100)|`cpu_time_delta / delta_t / n_cores` (all cores used fully = 100%)|`LocalMachine`|`Cgroup`|see below|
-|`memory_usage`|Gauge|Bytes|total pod's memory usage|`LocalMachine`|`Cgroup`|see below|
+|`memory_usage`|Gauge|Bytes|total cgroups's memory usage|`LocalMachine`|`Cgroup`|see below|
+|`memory_max`|Gauge|Bytes|Maximum memory available for the cgroup|`LocalMachine`|`Cgroup`|see below|
```
ahuret, line 18:
```suggestion
|`memory_usage`|Gauge|Bytes|total cgroups's memory usage|`LocalMachine`|`Cgroup`|[capped](#capped)|
|`memory_max`|Gauge|Bytes|Maximum memory available for the cgroup|`LocalMachine`|`Cgroup`|see below|
```

plugins/cgroups/raw/README.md
```diff
@@ -32,6 +33,18 @@ The **cpu** measurements have an additional attribute `kind`, which can be one o
...
+- `true`: the cgroup has a limit on this metric
+- `false`: the cgroup does not have a limit on this metric
+
+## Special Values for Maximum Metrics
+
+Some metrics that represent maximum limits (`memory_max`, `memory_swap_max`) are returned as `f64` instead of `u64` to handle a special case:
+
+- When these metrics return `-1`, it indicates that the capping is set to maximum (unlimited), meaning the actual limit value is not available or meaningful. This occurs when the cgroup has no explicit limit set and can use all available system resources.
+
+For all other values, these metrics represent the actual limit in bytes.
```
ahuret:
```suggestion
### Attributes

#### Capped

This attribute is set for metrics that have a relative `max` value (e.g. `memory_max`, `memory_swap_max`). The `capped` value is set to:
- `true` if the cgroup has an explicit limit on this metric
- `false` if the cgroup does not have a limit on this metric
- 

When `capped=false`, the value of the measurement will be set to `-1`.

ahuret (changes requested):

plugins/cgroups/raw/README.md
AngeCyp: Going to use **-1.0** as F64 and no attributes, not relevant anymore

plugins/cgroups/raw/README.md
AngeCyp: Going to use **-1.0** as F64 and no attributes, not relevant anymore

ahuret (approved):

PR #459, opened 2026-08-24
docs(amd-gpu-plugin): clarify safety comment for amd libary

Aymeric-Brochier:


ahuret (dismissed):
lgtm

plugins/amd-gpu/src/amd/device.rs
```diff
@@ -6,7 +6,10 @@ use std::collections::HashMap;
 
 use super::features::OptionalFeatures;
 
-/// SAFETY: The amd libary is thread-safe and returns pointers to a safe global state, which we can pass to other threads.
+/// SAFETY: The amd libary SHOULD BE thread-safe and returns pointers to a safe global state, which we can pass to other threads.
+/// BUT it IS NOT always the case (for example in old versions of the library or if built with thread safety mechanisms disabled).
+/// Since all GPUs are collected and polled by the same source in Alumet
+/// We can use unsafe here
```
AngeCyp: Maybe the use of uppercase is not relevant. 

> The AMD library generally provides thread-safe access to global state through its pointers. This is not guaranteed for all versions or configurations, such as older versions or builds with thread-safety mechanisms disabled.

> Alumet collects and polls all GPUs from the same source, ensuring that these pointers are not accessed concurrently.

WDYT ?

AngeCyp (changes requested):

plugins/amd-gpu/src/amd/device.rs
Aymeric-Brochier: "all GPUs are collected and polled by the same source" is not relevant for Send so I should not have mention it in this file

I clarified the reason for this decison in the README

TheElectronWill (approved):

AngeCyp (approved):

PR #465, opened 2026-08-27
fix(plugin-grace-hopper): handle modern kernel sysfs layout

ahuret:
Modern linux kernels organize differently some sysfs layout we parse in the grace-hopper plugin.

See: https://lkml.iu.edu/hypermail/linux/kernel/2503.2/04362.html

This PR enhance the explore function which make the parsing by looking in two different locations to make sure we support both old and moderns kernel versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5187,6 +5187,7 @@ dependencies = [
  "log",
  "pretty_assertions",
  "regex",
+ "serde",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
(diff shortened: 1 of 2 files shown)


TheElectronWill:
Can you include a link to the related LKML topic in the comments or commit description, so that we know where to get more info about the change?

Aymeric-Brochier (approved):
LGTM

ahuret:
> Can you include a link to the related LKML topic in the comments or commit description, so that we know where to get more info about the change?

for sure

TheElectronWill (approved):

AngeCyp (approved):

PR #467, opened 2026-08-31
Ci/add cargo semver checks

Aymeric-Brochier:
Do not merge yet : 
Testing ci workflow for semver (with a breaking change in plugin-csv ))

### DIFF
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -30,15 +30,44 @@ jobs:
         - uses: dtolnay/rust-toolchain@master
           with:
             toolchain: $CI_RUST_VERSION
-            components: rustfmt
-
         - uses: cargo-bins/cargo-binstall@main
 
         - name: Check dependencies
           run: |
            cargo binstall --no-confirm cargo-machete
            cargo machete
+  check_semver:
+      runs-on: ubuntu-latest
+      permissions:
+        issues: write
+        pull-requests: write
+      steps:
+        - uses: actions/checkout@v7
+          with:
+            fetch-depth: 0
+        - uses: dtolnay/rust-toolchain@master
+          with:
+            toolchain: $CI_RUST_VERSION
+        - uses: cargo-bins/cargo-binstall@main
 
+        - name: Check semver
+          id: check_semver
+          continue-on-error: true
+          run: |
+           cargo binstall --no-confirm cargo-semver-checks@0.45.0
+           # TODO remove -p plugin-csv after testing
+           cargo semver-checks -p plugin-csv --baseline-rev origin/${{ github.base_ref }}
+        - name: Add comment PR
+          if:  steps.check_semver.outcome == 'failure'
+          uses: actions/github-script@v9
+          with:
+            script: |
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: 'You have some breaking changes ! Please check the log of the check_semver job in the step check_semver before merging'
+              })
   build_and_test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/rust_ci.yml
```diff
@@ -30,15 +30,44 @@ jobs:
...
+        issues: write
+        pull-requests: write
+      steps:
+        - uses: actions/checkout@v7
+          with:
+            fetch-depth: 0
+        - uses: dtolnay/rust-toolchain@master
+          with:
+            toolchain: $CI_RUST_VERSION
+        - uses: cargo-bins/cargo-binstall@main
```
AngeCyp, line 51: IMO do as they say in the doc and pin the used version:
`bash
cargo tarpaulin \
  --package plugin-argos \
  --lib \
  collectors::system_collector::adapter::tests \
  -- --nocapture \
  --out Html
`

.github/workflows/rust_ci.yml
```diff
@@ -30,15 +30,44 @@ jobs:
...
 
+        - name: Check semver
+          id: check_semver
+          continue-on-error: true
+          run: |
+           cargo binstall --no-confirm cargo-semver-checks@0.45.0
+           # TODO remove -p plugin-csv after testing
```
AngeCyp, line 58: When would you remove it ?

plugins/amd-gpu/src/amd/device.rs
```diff
@@ -6,7 +6,10 @@ use std::collections::HashMap;
 
 use super::features::OptionalFeatures;
 
-/// SAFETY: The amd libary is thread-safe and returns pointers to a safe global state, which we can pass to other threads.
+/// SAFETY: The amd libary SHOULD BE thread-safe and returns pointers to a safe global state, which we can pass to other threads.
+/// BUT it IS NOT always the case (for example in old versions of the library or if built with thread safety mechanisms disabled).
+/// Since all GPUs are collected and polled by the same source in Alumet
+/// We can use unsafe here
```
AngeCyp, line 12: Should this be included in this PR ?

plugins/csv/src/lib.rs
```diff
@@ -42,7 +42,7 @@ impl AlumetPlugin for CsvPlugin {
             use_unit_display_name: self.config.use_unit_display_name,
             params: CsvParams {
                 delimiter: self.config.csv_delimiter,
-                late_delimiter: self.config.csv_late_delimiter,
+                late_delimiter: ',',
```
AngeCyp, line 45: We do not want always use "," do we ?

AngeCyp (changes requested):

TheElectronWill:
Now that it has been shown to work, you can revert the changes in plugin-csv.

.github/workflows/rust_ci.yml
Aymeric-Brochier: For testing purpose we only check the csv plugin ( else it would take too much time to check everything: around 15 min in the CI) 

I will work again on it next sprint with an improvment to detect which checks to do depending on file changed

plugins/amd-gpu/src/amd/device.rs
Aymeric-Brochier: No it was from another PR I will check and remove it

plugins/csv/src/lib.rs
Aymeric-Brochier: This is just for testing a breaking change in plugin-csv 

In the final PR it will be removed

Aymeric-Brochier:
There is a permission issues to write comment on the PR (On my fork it works).

I still need to keep the breaking change for now 
For the next step of testing : detecting the file changed and running checks only on plugin csv without specifying it 

I am converting this PR to draft for now

PR #470, opened 2026-09-11
feat(plugins/perf): support of system-wide counters

ahuret:
- support of system-wide counters
- target a single PMU
- automatically count of all PMU if no specific PMU configured (for hybrid cpu)

### DIFF
--- a/plugins/perf/README.md
+++ b/plugins/perf/README.md
@@ -25,10 +25,16 @@ suffix (and is normalized the same way); the metric then becomes `perf_{rename}`
 
 ### Attributes
 
+A native hardware/cache measurement carries a `pmu` attribute naming the core PMU it was counted on
+(`cpu_core`, `cpu_atom`, or `cpu`).
+
 Every measurement carries an `accuracy` attribute describing how faithful its value is (see
 [Counter multiplexing](#counter-multiplexing) below):
 
 - `exact`: an exact count.
+- `partial` (**only for hybrid CPU**): a per-PMU count on a hybrid CPU (see [Native events on hybrid CPUs](#native-events-on-hybrid-cpus)).
+  The event is pinned to one PMU while the process also runs on the others, so the counter was on
+  the PMU only part of the time. The value is **not** scaled.
 - `extrapolated`: the value includes at least one multiplexed interval that was extrapolated (only
   happens when `multiplexing_auto_scale` is on). It is an estimate, which may be slightly above or
   below the truth.
@@ -36,9 +42,10 @@ Every measurement carries an `accuracy` attribute describing how faithful its va
   reported raw (`multiplexing_auto_scale` off) or because the counter was starved for some intervals
   (the events could not be counted at all).
 
-Because the reported values are cumulative counters, the accuracy only ever degrades from `exact` to
-`extrapolated` to `underestimated`, and never improves: a single imperfect interval affects every
-value reported afterwards.
+Because the reported values are cumulative counters, the accuracy only ever degrades and never
+improves: a single imperfect interval affects every value reported afterwards. `partial` applies only
+to PMU-pinned events (a hybrid-CPU artifact) and is never scaled; `extrapolated`/`underestimated`
+describe genuine multiplexing on a single-PMU event.
 
 ## Configuration
 
@@ -122,18 +129,25 @@ Each event is a string of the form:
 
 #### Event formats
 
-The `<event>` part can be one of three forms:
+The `<event>` part can be one of these forms:
 
 - **native** : a symbolic event name (e.g. `INSTRUCTIONS`, `LL_READ_MISS`), resolved against the
-  built-in kernel tables listed in [Native event names](#native-event-names). (**Supported**)
+  built-in kernel tables listed in [Native event names](#native-event-names).
+- **native on a PMU** : a native name pinned to a specific PMU, `pmu/NAME` (e.g.
+  `cpu_core/INSTRUCTIONS`, `cpu_atom/CACHE_MISSES`).
 - **libpfm** : any other name, optionally with unit masks (e.g. `RESOURCE_STALLS:ANY`,
-  `MEM_LOAD_RETIRED:L3_MISS`), resolved through [libpfm](#libpfm-events). (**Supported**, requires libpfm)
+  `MEM_LOAD_RETIRED:L3_MISS`), resolved through [libpfm](#libpfm-events). (requires libpfm)
 - **raw-hex** : a raw code `rN`, where `N` is a hexadecimal value representing the raw register
   encoding, with the layout described by `/sys/bus/event_source/devices/<pmu>/format/*`. It targets
-  the default raw PMU. (**Supported**, see [Raw events](#raw-events))
+  the default raw PMU. (see [Raw events](#raw-events))
+- **raw on a PMU** : `pmu/rN`, the same raw code but on a named PMU (addressed by its sysfs `type`),
+  e.g. `uncore_imc_0/r0x1`. This is the only way to reach a PMU that has no generic namespace, such
+  as the uncore, `power` (RAPL), or `cstate_*` PMUs. (see [Raw events](#raw-events)
+  and [System-wide events](#system-wide-events-uncore-rapl-cstate))
 
 Any event may be followed by `#` and a list of [modifiers](#modifiers), e.g.
-`INSTRUCTIONS#u` or `CACHE_MISSES#u:k`.
+`INSTRUCTIONS#u` or `CACHE_MISSES#u:k`. Modifiers do not apply to
+[system-wide events](#system-wide-events-uncore-rapl-cstate) and are rejected on them.
 
 #### Native event names
 
@@ -155,6 +169,20 @@ For example: `LL_READ_MISS`.
 To learn more about the standard events, please refer to the [`perf_event_open` manual](https://man7.org/linux/man-pages/man2/perf_event_open.2.html).
 To list the events that are available on your machine, run the `perf list` command.
 Note that based on your kernel version, some events could be unavailable.
+
+#### Native events on hybrid CPUs
+
+On hybrid CPUs (e.g. Intel P-core/E-core, exposed as the `cpu_core` and `cpu_atom` PMUs), a plain
+generic hardware/cache event is counted on **every** PMU automatically.
+They share the metric name (`perf_instructions`) and are told apart by their `pmu` attribute (`cpu_core` / `cpu_atom`);
+
+To measure one PMU only, qualify the name with the PMU: `cpu_core/INSTRUCTIONS` or
+`cpu_atom/INSTRUCTIONS`. This works for hardware and cache events; **software** events (e.g.
+`CONTEXT_SWITCHES`) are CPU-wide and have no PMU, so pinning them is rejected.
+
+These are per-process/cgroup events (the core PMUs are task-attachable), so they follow the observed
+entity like any native event. Their `resource` is the CPU package the PMU sits in
+(`cpu_package`), and the `pmu` attribute is added.
 
 #### Libpfm events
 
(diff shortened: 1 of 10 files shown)


plugins/perf/src/cpu.rs
```diff
@@ -31,6 +31,15 @@ fn parse_cpu_list(cpulist: &str) -> anyhow::Result<Vec<u32>> {
     Ok(cpus)
 }
 
+/// Read the physical package (socket) id a CPU belongs to, from sysfs topology.
+pub(crate) fn package_of(cpu: u32) -> anyhow::Result<u32> {
+    let path = format!("/sys/devices/system/cpu/cpu{cpu}/topology/physical_package_id");
```
Aymeric-Brochier, line 39: This seems hardware dependant : 
Do you assume it is always a physical socket number ?
```
physical_package_id:

physical package id of cpuX. Typically corresponds to a physical socket number, but the actual value is architecture and platform dependent.
```

https://www.infradead.org/~mchehab/kernel_docs/admin-guide/cputopology.html

Would be interesting to be sure is ok on all the hardware we support

plugins/perf/src/native.rs
```diff
@@ -191,4 +213,29 @@ mod tests {
     fn unknown_name_is_rejected() {
         assert!(parse("DEFINITELY_NOT_A_REAL_EVENT_XYZ").is_err());
     }
+
+    #[test]
+    fn extended_config_packs_pmu_type_high() {
+        // (pmu_type << 32) | config. e.g. cpu_core (type 4) + generic INSTRUCTIONS (1).
```
Aymeric-Brochier: should be  config | (pmu_type << 32) ... 
because of the implementation of the function extended_config (l60)
(even if it the result is the same)

plugins/perf/src/raw.rs
```diff
@@ -27,17 +38,21 @@ fn raw_config(token: &str) -> Option<u64> {
     u64::from_str_radix(digits, 16).ok()
 }
 
-fn build(config: u64, original: &str) -> NamedPerfEvent {
-    NamedPerfEvent {
+fn build(pmu: Option<&str>, config: u64, original: &str) -> anyhow::Result<NamedPerfEvent> {
```
Aymeric-Brochier, line 41: build function is defined after it's usage in parse function 

Maybe it could be moved before parse.

plugins/perf/src/raw.rs
Aymeric-Brochier, line 32: raw_config function is defined afters it's usage in parse 

Maybe it could be moved before parse.

plugins/perf/src/lib.rs
```diff
@@ -80,10 +80,20 @@ impl AlumetPlugin for PerfPlugin {
 
         let mut config = self.config.lock().unwrap();
 
+        // Events fanned onto several core PMUs share one metric (same name, told apart by their
+        // `pmu` attribute), so a metric is created once per distinct name and reused.
         let mut metrics = Vec::with_capacity(config.events.len());
+        let mut by_name: std::collections::HashMap<String, TypedMetricId<u64>> = std::collections::HashMap::new();
```
TheElectronWill, line 86: import HashMap for style consistency

plugins/perf/src/native.rs
```diff
@@ -39,6 +40,27 @@ pub fn parse(name: &str) -> anyhow::Result<NamedPerfEvent> {
...
 
+/// Pin an already-resolved native event to a PMU of numeric type `pmu_type`.
+/// Only hardware and hardware-cache events can be pinned.
+pub(crate) fn pin(base: &NamedPerfEvent, pmu: &str, pmu_type: u32) -> anyhow::Result<EventEncoding> {
+    match base.encoding.type_ {
+        PERF_TYPE_HARDWARE | PERF_TYPE_HW_CACHE => Ok(EventEncoding {
+            config: extended_config(base.encoding.config, pmu_type),
+            ..base.encoding
+        }),
+        PERF_TYPE_SOFTWARE => anyhow::bail!(
```
TheElectronWill, line 51: IMO it would be more clear to use `Err` explicitly here, it would make the code more regular (and reduce compile time a liiiiitle bit).
Usually, `bail!` is reserved to situations where you have to return early from a deeply nested block of code.

plugins/perf/README.md
```diff
@@ -25,20 +25,27 @@ suffix (and is normalized the same way); the metric then becomes `perf_{rename}`
...
 ### Attributes
 
+A native hardware/cache measurement carries a `pmu` attribute naming the core PMU it was counted on
+(`cpu_core`, `cpu_atom`, or `cpu`).
+
 Every measurement carries an `accuracy` attribute describing how faithful its value is (see
 [Counter multiplexing](#counter-multiplexing) below):
 
 - `exact`: an exact count.
+- `partial` (**only for hybrid CPU**): a per-PMU count on a hybrid CPU (see [Native events on hybrid CPUs](#native-events-on-hybrid-cpus)).
```
TheElectronWill, line 35: What happens when a partial counter is _also_ multiplexed?
Shouldn't we decouple this case from the multiplexing case?

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
-
     // Fall back to libpfm.
-    pfm::encode(name)
-        .with_context(|| format!("unknown event '{name}': not a native event, and libpfm could not encode it"))
+    let base = pfm::encode(name)
+        .with_context(|| format!("unknown event '{name}': not a native event, and libpfm could not encode it"))?;
+    Ok((base, Placement::Unpinned))
+}
+
+struct CoreTarget {
```
TheElectronWill, line 371: What is a CoreTarget? (missing doc)

plugins/perf/src/spec.rs
```diff
@@ -179,12 +194,28 @@ pub struct ConfiguredEvent {
     modifiers: Modifiers,
 }
 
+/// How an event must be opened, decided by the PMU it targets.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Scope {
+    TaskAttached { binding: Option<CoreBinding> },
```
TheElectronWill, line 200: A small doc on each variant would help the reader.
By "task", do you mean a kernel task? Or is it a "process or cgroup"? (if the latter, perhaps another name would be more suitable - just `Attached`?)

plugins/perf/src/spec.rs
```diff
@@ -80,6 +91,10 @@ impl Modifiers {
         Ok(m)
     }
 
+    fn any(&self) -> bool {
```
TheElectronWill, line 94:
```suggestion
    fn has_any(&self) -> bool {
```

plugins/perf/src/source.rs
```diff
@@ -282,3 +235,192 @@ impl PerfEventSourceBuilder {
...
+                .observe_cgroup(fd)
+                .one_cpu(cpu_id.expect("a cgroup group is always bound to a specific cpu") as usize);
+        }
+    }
+}
+
+/// Add `event` to the group matching `(cpu_id, key)`, creating that group (a fresh `DUMMY` leader)
+/// if none exists yet.
+#[allow(clippy::too_many_arguments)]
+fn ensure_group_and_add(
```
TheElectronWill, line 268: Ouch :smile:
Perhaps it would help to have a structure dedicated to managing a list of EventGroup + a kind of builder pattern, or pass some parameters as a struct so they are named? I know the function isn't used in 15 different places, but there's probably still room for improvement ^^

plugins/perf/src/pmu.rs
```diff
@@ -0,0 +1,178 @@
...
+pub fn core_pmus() -> anyhow::Result<Vec<CorePmu>> {
+    let mut out = Vec::new();
+    let entries = match fs::read_dir("/sys/bus/event_source/devices") {
+        Ok(e) => e,
+        Err(_) => return Ok(out),
+    };
+    for entry in entries.flatten() {
+        let name = entry.file_name().to_string_lossy().into_owned();
+        // A core PMU has a `cpus` list and no `cpumask` (that would make it system-wide).
+        if !matches!(read_cpumask(&name), Ok(None)) {
```
TheElectronWill, line 75:
```suggestion
        let Ok(None) = read_cpumask(&name) else { continue };
```
for consistency and ease of reading

TheElectronWill (changes requested):

plugins/perf/src/pmu.rs
```diff
@@ -0,0 +1,178 @@
...
+pub fn read_cpumask(pmu: &str) -> anyhow::Result<Option<Vec<u32>>> {
+    read_cpu_list_file(pmu, "cpumask")
+}
+
+/// Read a PMU's `cpus`.
+pub fn read_cpus(pmu: &str) -> anyhow::Result<Option<Vec<u32>>> {
+    read_cpu_list_file(pmu, "cpus")
+}
+
+fn read_cpu_list_file(pmu: &str, file: &str) -> anyhow::Result<Option<Vec<u32>>> {
```
Aymeric-Brochier, line 45: Move `read_cpu_list_file` before `read_cpus` and `read_cpumask` since these fuctions use it

plugins/perf/src/pmu.rs
```diff
@@ -0,0 +1,178 @@
...
+            cpus,
+            package,
+        });
+    }
+    out.sort_by(|a, b| a.name.cmp(&b.name));
+    Ok(out)
+}
+
+/// The package shared by every CPU in `cpus`, or `None` if they span several packages
+pub fn single_package(cpus: &[u32]) -> Option<u32> {
```
Aymeric-Brochier, line 91: Move `single_package()` above `core_pmus()` since it is used in it

plugins/perf/src/pmu.rs
```diff
@@ -0,0 +1,178 @@
...
+        let Some(pmu) = first_pmu_with_cpumask() else {
+            eprintln!("skipping cpumask_present_reads_as_some: no system-wide PMU found in sysfs");
+            return;
+        };
+        let cpus = read_cpumask(&pmu).unwrap().expect("cpumask file exists");
+        assert!(!cpus.is_empty(), "cpumask of {pmu} should list at least one CPU");
+    }
+
+    /// Find any PMU exposing a `cpumask` in sysfs (a system-wide PMU), returning its name.
+    fn first_pmu_with_cpumask() -> Option<String> {
```
Aymeric-Brochier, line 154: Move `first_pmu_with_cpumask()` before `cpumask_present_reads_as_some()` since it is used int it

plugins/perf/src/raw.rs
```diff
@@ -79,6 +94,42 @@ mod tests {
...
+    #[test]
+    fn pmu_raw_with_unknown_pmu_errors() {
+        // The `pmu/rN` shape matched, but the PMU does not exist: this must surface as an error,
+        // not as "not recognised".
+        let result = parse("definitely_not_a_pmu_xyz/r0x1").expect("recognised as pmu/rN");
+        assert!(result.is_err());
+    }
+
+    /// Find any PMU exposing a numeric `type` in sysfs, returning its name and type.
+    fn first_available_pmu() -> Option<(String, u32)> {
```
Aymeric-Brochier, line 125: Move `first_available_pmu()` before  `pmu_raw_uses_the_pmu_type()` since used in it

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
 
-    Ok(ParsedEvent {
-        metric_suffix: sanitize(metric_suffix),
-        description: resolved.description,
-        event: ConfiguredEvent {
-            encoding: resolved.encoding,
-            modifiers,
-        },
-    })
+fn detect_system_wide(name: &str) -> anyhow::Result<Option<(String, Vec<u32>)>> {
```
Aymeric-Brochier, line 286: Move `detect_system_wide(name: &str)` before `parse()` since it is used in it

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
+    Unpinned,
+    /// An explicit core PMU.
+    PmuPinned(String),
+    /// A bare native hardware/cache event, to be counted on every core PMU.
+    CoreGeneric,
+}
+
+/// Resolve an event name (no modifiers) into its encoding and how it places onto the core PMU(s).
+/// Each encoder builds the [`NamedPerfEvent`] in its own module; this dispatches to them.
+fn resolve_base(name: &str) -> anyhow::Result<(NamedPerfEvent, Placement)> {
```
Aymeric-Brochier, line 335: Move `resolve_base(name: &str)` before `parse()` since it is used in it

Maybe more difficult to move since it would need to stay close to enum `Placement` 

Maybe move both or leave like this

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
+    }
 }
 
-/// Resolve an event name (no modifiers) into a [`NamedPerfEvent`]. Each encoder builds the `NamedPerfEvent`
-/// in its own module; this only dispatches to them. See the module docs for the recognised forms.
-fn resolve_event(name: &str) -> anyhow::Result<NamedPerfEvent> {
-    // raw-hex route: `rN`, a raw code on the default raw PMU.
-    if let Some(result) = raw::parse(name) {
-        return result;
+fn resolve_task_attached(name: &str, rename: Option<&str>, modifiers: Modifiers) -> anyhow::Result<Vec<ParsedEvent>> {
```
Aymeric-Brochier, line 293: Move `resolve_task_attached(...)` before `parse()` since it is used in it

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
+        .with_context(|| format!("unknown event '{name}': not a native event, and libpfm could not encode it"))?;
+    Ok((base, Placement::Unpinned))
+}
+
+struct CoreTarget {
+    pmu_type: Option<u32>,
+    binding: CoreBinding,
+}
+
+fn core_targets() -> anyhow::Result<Vec<CoreTarget>> {
```
Aymeric-Brochier, line 384: Move closer to `resolve_task_attached(..)` if possible

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
+            binding: CoreBinding {
+                pmu: c.name,
+                resource: package_resource(c.package),
+                cpus: c.cpus,
+            },
+        })
+        .collect())
+}
+
+fn core_binding(pmu: &str) -> anyhow::Result<CoreBinding> {
```
Aymeric-Brochier, line 411: Move closer to `resolve_task_attached(..)` if possible

plugins/perf/src/spec.rs
```diff
@@ -203,56 +248,176 @@ impl ConfiguredEvent {
...
+    };
+    let resource = package_resource(pmu::single_package(&cpus));
+    Ok(CoreBinding {
+        pmu: pmu.to_owned(),
+        cpus,
+        resource,
+    })
+}
+
+fn package_resource(package: Option<u32>) -> Resource {
```
Aymeric-Brochier, line 424: Move above it's usage if possible

plugins/perf/README.md
ahuret: Its value is left raw, reported exactly as the kernel gives it, never scaled. The **partial** attribute warns the user that the value covers only one PMU and that we don't know if it was multiplexed or not.

On a hybrid CPU we open one counter per PMU. For such a counter, **running** time means "how long I was measured on this PMU", while **enabled** time means "how long my task ran in total". So **running** is normally lower than **enabled**, which, for an ordinary counter, is exactly what triggers multiplexing, but here usually means something else: the task simply ran on another PMU (or maybe really multiplexed, we don't know...). Treating that as multiplexing would scale both PMU counters against a total that includes the other PMU time, inflating the values.

However, I had an idea: summing up the **running** time of both PMU, if the sum is equal to the **enabled** time, this means it didn't multiplexed. This way we can distinguish a multiplexing case from an exact case.
However I made a try but it doesn't work: the two counters are read one just after the other, not at the same moment. So the two sides are always off by a tiny amount (a few microseconds), so we can never know if it has multiplexed or not.

plugins/perf/src/source.rs
ahuret: I've created an intermediate struct called GroupSpec that manage almost all redundant parameters. So we have way less paramters. However I didn't introduced yet the builder, it doesn't seems necessary now since it's not used a lot,  the system of injected mutable vec is quite simple. tell me if you really believe I should introduce that builder now

plugins/perf/src/cpu.rs
ahuret: we just use it as an ID to regroup CPU belonging to the same package together, so having the value dependent on the architecture and on the platform doesn't change the meaning of it from our point of view. I added a comment to explain it more deeply on top of the function itself.

ahuret:
@TheElectronWill  @Aymeric-Brochier pushed a new commit to easy your new review: https://github.com/alumet-dev/alumet/pull/470/changes/ecdf02bdc1ed230b911dc74bc2560a49657d9259

